PR #13, opened 2016-01-14
Price per subscription option

seb86:
I have added the option to enter a price per subscription option.

Here are some screenshots to show the changes. https://cloudup.com/calXMRYWl-m

I look forward to your feedback.

### DIFF
--- a/assets/css/wcsatt-write-panels.css
+++ b/assets/css/wcsatt-write-panels.css
@@ -64,6 +64,22 @@
 	content: " ";
 }
 
+#wcsatt_data table.subscription_scheme_table p._subscription_price_field {
+  min-width: 198px !important;
+}
+
+#wcsatt_data table.subscription_scheme_table p._subscription_price_field label {
+  display: initial;
+  line-height: 2.3em;
+  width: initial;
+  margin: 0 0.8em 0 0 !important;
+}
+
+#wcsatt_data table.subscription_scheme_table p._subscription_price_field input,
+#wcsatt_data table.subscription_scheme_table .wc-metaboxes-wrapper .wc-metabox table td p._subscription_price_field input {
+  min-width: 1em;
+}
+
 /*
  * Admin settings
  */
--- a/changelog.txt
+++ b/changelog.txt
@@ -1,4 +1,6 @@
 *** WooCommerce Subscribe All the Things ***
+2016.01.14 - version 1.0.3
+ * Added - Each subscription option can have it's own price.
 
 2015.11.27 - version 1.0.2
  * Fix - Minified scripts not in repo.
--- a/includes/admin/class-wcsatt-admin.php
+++ b/includes/admin/class-wcsatt-admin.php
@@ -62,6 +62,7 @@ class WCS_ATT_Admin {
 		if ( ! $subscription_schemes ) {
 			$subscription_schemes = array(
 				array(
+          'subscription_price'           => 0,
 					'subscription_period_interval' => 1,
 					'subscription_period'          => 'month',
 					'subscription_length'          => 0,
@@ -169,7 +170,7 @@ class WCS_ATT_Admin {
 
 				foreach ( $posted_schemes as $posted_scheme ) {
 
-					$scheme_id = $posted_scheme[ 'subscription_period_interval' ] . '_' . $posted_scheme[ 'subscription_period' ] . '_' . $posted_scheme[ 'subscription_length' ];
+					$scheme_id = $posted_scheme[ 'subscription_period_interval' ] . '_' . $posted_scheme[ 'subscription_period' ] . '_' . $posted_scheme[ 'subscription_length' ] . '_' . $posted_scheme[ 'subscription_price' ];
 
 					if ( in_array( $scheme_id, $scheme_ids ) ) {
 						continue;
@@ -246,14 +247,32 @@ class WCS_ATT_Admin {
 		}
 
 		if ( ! empty( $scheme_data ) ) {
+			$subscription_price           = $scheme_data[ 'subscription_price' ];
 			$subscription_period          = $scheme_data[ 'subscription_period' ];
 			$subscription_period_interval = $scheme_data[ 'subscription_period_interval' ];
 			$subscription_length          = $scheme_data[ 'subscription_length' ];
 		} else {
+			$subscription_price           = '';
 			$subscription_period          = 'month';
 			$subscription_period_interval = '';
 			$subscription_length          = '';
 		}
+
+		// Subscription Price
+		woocommerce_wp_text_input( array(
+			'id'          => '_subscription_price',
+			'class'       => 'wc_input_subscription_price',
+			// translators: %s is a currency symbol / code
+			'label'       => sprintf( __( 'Subscription Price (%s)', 'woocommerce-subscriptions' ), get_woocommerce_currency_symbol() ),
+			'placeholder' => _x( 'e.g. 5.90', 'example price', 'woocommerce-subscriptions' ),
+			'type'        => 'text',
+			'custom_attributes' => array(
+					'step' => 'any',
+					'min'  => '0',
+			),
+			'name'        => 'wcsatt_schemes[' . $index . '][subscription_price]',
+			'value'       => $subscription_price
+		) );
 
 		// Subscription Period Interval
 		woocommerce_wp_select( array(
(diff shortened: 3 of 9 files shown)


titocalabrese:
That's interesting! 
It would be great to add also the possibility to discount of a percentage the subscription based on the items that are part of the subscription itself (like Amazon does). Your feature fits well with simple products but it's not applicable in case of composite product that usually have optional component (i.e. in this case it is not possible to predict the items that the user will choose and therefore the total cost of the subscription

franticpsyx:
@seb86 thanks for the PR :+1: 
@titocalabrese The defined scheme price **could** override the base price of the composite container, when the composite is priced per-item. I'll need to take a closer look to confirm this, though.

I will give this a better look and get back to you asap :)

titocalabrese:
@franticpsyx you are right! The override is possible for the composite container. Indeed if you have optional components you are not able to forecast a price for the container. With percentage discount it is possible to overcome this limitation :)

seb86:
It's a start in the right direction.

franticpsyx:
>It's a start in the right direction.

>With percentage discount it is possible to overcome this limitation :)

I think I would prefer a Discount field, since that would allow the defined price override to be applicable to Bundles/Composites, as well - an major motivation for coding this plugin in the first place.

On the other hand, discounts sometimes make it hard to get an exact price.

- Discount-based overrides sacrifice some usability/precision, while 
- Price-based overrides sacrifice some functionality.

I'd take (1) over (2). Not sure if we should have both. Thoughts? @jconroy ;)

seb86:
@franticpsyx I think we could add both depending on the type of product that is created.

seb86:
@franticpsyx A simple product could have the option to choose either price or discount while other types of products would only have the one option.

seb86:
@franticpsyx Would the discount be a flat-rate or a percentage of the regular price?

franticpsyx:
>Would the discount be a flat-rate or a percentage of the regular price?

Percentage off the price of the product, so that it can propagate down to any dependent (bundled/composited/...) cart items. We could use a filter to determine whether the % discount is calculated off the regular or the sale price (if it exists).

franticpsyx:
@seb86 a quick update about this - we are planning to work on some SATT-related tasks in a few days, and this is definitely on my to-do list :)

seb86:
@franticpsyx Great. Let me know if you need any help.

furchildpets:
@seb86 @franticpsyx We think that this is a great feature!! It would be perfect for our needs. Let me know if you need any beta testing.

We like your idea as well about the discount being a flat-rate or % off the regular price.  Since the regular price will be multiplied by the number of weeks, months, days, etc.. of the subscription it could be applied to any length of time and provide accurate prices.

We look forward to your work.

furchildpets:
@seb86 @franticpsyx Also can you add a div class to the Subscription Prompt text box so that I can be customized with css?

Thanks

seb86:
@furchildpets Which prompt text box are you referring to?

furchildpets:
@seb86 The Subscription Prompt field provided. See the attached image...  I learned that you can add html there so I created my own div class, but it would be better if was in the page template.

image

I have done so css customization and this is the result...

image

seb86:
@furchildpets Just added two action hooks that you can use to customize the subscription prompt area. Hope that helps.

furchildpets:
@seb86 Ok great!!! Thanks

furchildpets:
Hello @seb86 @franticpsyx You probably know this already but wanted to let you know that the subscription prices add on to the fixed / base price of the composite product instead of overriding the Total Price.

You can see an example here....

image


Also when you build your composite product and price products individually the subscription price adds on to those as well. So no matter what you put in your Subscription Prices they do not override the Total Price field of the composite product.

This leads me to think that should be two modes for the Subscription Prices;

1. To be an additional cost on top of the base / fixed price or total of individual prices of the single products.

2. Or override / discount the fixed base price of the composite product.

Well, once again let me know if I can be of any help - beta testing, etc...

Thanks

Paul

seb86:
@furchildpets Thank you for your feedback. We will look into this.

louis2b:
Hi everyone, as a shop owner that mini extension is exactly what I am hoping to do with subs. But as you mentioned being able to change the price for each subs scheme is vital. 

I understand the advantage of % discount but I think an option to obtain a clean round price is important (either with a fixed price that overrides, or better a fixed discount (compatible with bundles and composites))

franticpsyx:
@seb86 Can you have a look at the https://github.com/Prospress/woocommerce-subscribe-to-all-the-things/tree/issue-13-patches branch?

@louis2b We will do both. See https://github.com/Prospress/woocommerce-subscribe-to-all-the-things/tree/issue-13-patches and let us know if you find any issues :)

Haven't tested it thoroughly with Bundles + Composites under all price-overriding scenarios, so if you notice anything strange let me know. Thanks!

seb86:
@franticpsyx I will try to find some time over the weekend but I will need access to your bundles and composites plugin repositories. Would you be able to do that?

louis2b:
Hi @franticpsyx thanks for the update, I have tested the extension with bundle products, I noticed a few problems. 

- when "per item pricing" enabled and subs price set to "inherit from product price" there are display issues on the page (but everything is calculated correctly in the cart). To be more precise there are 4 places where the price is displayed. (1) "from price" on the top below product name with minimum possible value (depending on min quant for bundle products), (2) the radio button with subs schemes also show the "minimum price", but do not take into account the %discount (which is then correctly calculated in cart) (3) a total price (linked to bundle pdt) next to the add to cart button, updated with sum of all selected bundle products but not updated when a subs scheme is selected (4) a last total price below the add to cart button, showing minimum price and never updated (same as (1)). 
My opinion is that (3) and (4) should be merged and that the result plus (2) should be updated dynamically with selection (product number + subs scheme). But I dont know how easy it would be to have the radio button updated when increasing/decreasing the number of bundled products. 

- when "per item pricing" enabled and subs price set to "override product price" (although I believe those 2 choices are contradictory) the same display issue persist plus an anormal price in checkout. To be more precise again, (1) (3) and (4) have the same behavior than previously. (2) still shows minimum possible price instead of overriden prices.  In cart the sub price is correctly displayed (overriden prices) but the total is higher and does not correspond to any price. It might be a sum of several of them. 

Hope that helps ! Let me know if I was not clear 
BEst

furchildpets:
@franticpsyx Thanks for the updated release and pointing me in the right direction. We really appreciate your continuing work on this extension.  

I have tested this with our dev site with Product Bundles and it is working better, but there is one BIG problem, that I believe @louis2b has seen as well...

The correct price and total go into the cart page, but not into the Cart Widget in the Header. It has the original product price when you add any subscription value. After you refresh the cart page or go to another site page the cart widget price changes and reflects the subscription override price.

I have turned off woocommerce > settings > ajax cart updates and the same thing happens.

Before adding item to the cart... Product Bundle with original price....

image

Override Product - 2 Subscription levels - one with sale price...

image

Display on Product page...

image

Subscription Product added to Cart. Cart Widget price updates with Orginal Price not Subscription price....

image

However Cart Totals are correct...

image

When the page is refreshed or you go to another site page the Cart Widget reflects the correct price...

image

Our Cart Widget is part of the standard Woocommerce integration into the Mill Theme.  No customization there.

Any ideas why it is not reflecting the correct price when the product is first added?

Thanks,

P

franticpsyx:
@louis2b @furchildpets Thanks for the feedback - there is still some work to do on that branch, for instance related to the mini-cart issue. A couple other issues will be harder to fix - for instance, Bundles compatibility with discounts / price overrides and per-item priced Bundles will take some time but it's definitely on my to-do list. :+1:

furchildpets:
@franticpsyx Your welcome -- for us if we can just get the mini-cart issue solved then we have enough to go live.  All of our product bundles are set prices and the override product price is working well.

Something interesting to note... when we activate the WooCommerce Cart Tab plugin --
the Cart Tab has the original product bundle price as the page first loads, but then updates automatically to the correct subscription price 1 sec later.

Plugin LInk - https://wordpress.org/plugins/woocommerce-cart-tab/

image

As a work around I am thinking about forcing the visitor to the Cart page after adding a product and then hiding the Cart value via css display:none; on just that specific page. When they go to any other site page the Mini Cart reflects the correct price total.

The bundles compatibility with price overrides - reg. and sale are working well for us.

Thanks again.

franticpsyx:
https://github.com/Prospress/woocommerce-subscribe-to-all-the-things/commit/64b50fb4b7a6d85b3df46257dd9042e8fc9763b8 should resolve the mini-cart issue.

Re the issues with Product Bundles and discounts, tackling them is possible but will require a significant amount of work. At this time, it is difficult to give a precise ETA.

Closing this - please continue the discussion at https://github.com/Prospress/woocommerce-subscribe-to-all-the-things/issues/18

Issue #30, opened 2016-04-21
Simple Product on Sale with Subscription Options


seb86:
@thenbrent and @franticpsyx I noticed that if a simple product that is on sale has subscription options, the price on the subscriptions are still taking a percentage of the regular price not the sale price.

![Simple Product on Sale with Subscription Options](https://cldup.com/QrsyTxoCOg.png)

Should it not inherit the sale price of the product when it is on sale and provide a discount percentage from that?

franticpsyx:
See https://github.com/Prospress/woocommerce-subscribe-to-all-the-things/blob/master/includes/class-wcsatt-schemes.php#L159

The filter can be used to apply the percentage on top of the sale price :)

seb86:
@franticpsyx OK but the average user who does not know how to code is not going to know that filter is available so should we not then add a new option to deterin the result of that filter?

If not then at least put some kind of message in the product panel to notify the user what happens. What do you think @thenbrent ?

thenbrent:
@franticpsyx what was the thinking behind using the regular price by default instead of the sale price?

I'm inclined to agree with @seb86 that the default behaviour should be to honour the `$price` value so that sale prices and other filters applied to the product's price are honoured before applying the SATT discount.

franticpsyx:
>what was the thinking behind using the regular price by default instead of the sale price

It was based on user feedback from the Product Bundles extension, where (oddly) I had quite a few requests to have discounts applied on top of regular prices to avoid sale prices propagating into bundles.

Similarly in this case, my thinking was that people may temporarily want to have a sale price applied for a limited time when doing a one-off purchase, without affecting the subscription scheme prices (always based on the regular price).

thenbrent:
Hrm, yeah I thought it may have been related to Bundles where it makes sense... the second use case of temporary sales also makes sense.

I think we can probably preserve the current behaviour and instead of changing it, try to avoid the confusion by being explicit about what will be discounted by changing the `Discount %` label to say `Discount Regular Price (%)` (e.g. https://cloudup.com/cfnw7hrOGET)

PR #32, opened 2016-04-26
Support for variable products

seb86:
It is not complete just yet but its a start in the right direction. There are things still incomplete and buggy but I thought I share what I have so far.

I have also improved the user interface. All fields are now aligned properly. Upon pressing the "Remove" button on a subscription option, the user is now notified with a dialog to confirm.


seb86:
@thenbrent @franticpsyx Let me know what you think so far once you get a chance to take a look.

franticpsyx:
@seb86 - Heym thanks for all the hard work! I send some comments once I review your PR, but in general I would advise against developing major new features like this one without having a general consensus in place (or even a basic spec).

I recently gave some thought to @thenbrent 's comments in https://github.com/Prospress/woocommerce-subscribe-to-all-the-things/issues/2 and looked at some of the feedback I've received over time. I was starting to lean towards Brent's opinion to start with variable product support at product level to see what kind of challenges we'll face in the front-end before we go all the way to variation level handling.

Here it looks like you have worked hard on most of the admin area requirements to support sub schemes at variation level, leaving the front-end part for later (which I'm fine with).

However I'm not yet 100% positive that we should start with variation-level sub schemes support.

@thenbrent ?

seb86:
@franticpsyx This is actually what I have done so far for a client. The first PR I did for this project was also from the same client. He really likes this extension and wanted to help with its development so that he could use it for his store.

I look forward to your feedback.

seb86:
Had a bit of an issue with undoing a commit but I think it's now back on track.

includes/class-wcsatt-display.php
```diff
@@ -103,11 +112,10 @@ public static function convert_to_sub_product_options() {
 			$product_type = 'simple';
 		}
 
-		if ( $product_type != 'variable' ) {
-			$post_id = $product->id;
-		} else {
-			$post_id = $product->children['visible']['0']; // Gets the first enabled variation
-		}
+		// If the product is a variable product then dont display the subscription options
+		if ( $product_type == 'variable' ) return false;
```
seb86: The reason for doing this is because the variation products need to display the subscription options via Javascript which needs to be appended to the single variation wrap without modifying the template files.

includes/class-wcsatt-display.php
```diff
@@ -72,12 +75,12 @@ public static function frontend_scripts() {
...
 		$suffix = defined( 'SCRIPT_DEBUG' ) && SCRIPT_DEBUG ? '' : '.min';
 
-		wp_register_style( 'wcsatt-css', WCS_ATT()->plugin_url() . '/assets/css/wcsatt-frontend.css', false, WCS_ATT::VERSION, 'all' );
+		wp_register_style( 'wcsatt-css', WCS_ATT()->plugin_url() . '/assets/css/wcsatt-frontend' . $suffix . '.css', false, WCS_ATT::VERSION, 'all' );
 		wp_enqueue_style( 'wcsatt-css' );
 
 		// Product Page
 		if ( is_product() ) {
-			wp_register_script( 'wcsatt-add-to-cart-variation', WCS_ATT()->plugin_url() . '/assets/js/wcsatt-add-to-cart-variation.js', array( 'jquery', 'wc-util' ), WCS_ATT::VERSION, true );
+			wp_register_script( 'wcsatt-add-to-cart-variation', WCS_ATT()->plugin_url() . '/assets/js/wcsatt-add-to-cart-variation' . $suffix . '.js', array( 'jquery', 'wp-util' ), WCS_ATT::VERSION, true );
```
seb86: > Note: Will need to minify the assets for the frontend once ready.

assets/js/wcsatt-add-to-cart-variation.js
```diff
@@ -0,0 +1,20 @@
...
+	var form                  = $(this),
+			single_variation      = form.find( '.single_variation' ),
+			product               = form.closest( '.product' ),
+			product_id            = parseInt( form.data( 'product_id' ), 10 ),
+			product_variations    = form.data( 'product_variations' ),
+			reset_variations      = form.find( '.reset_variations' ),
+			template              = wp.template( 'variation-template' ),
+			unavailable_template  = wp.template( 'unavailable-variation-template' ),
+			single_variation_wrap = form.find( '.single_variation_wrap' );
+
```
seb86: Not sure yet if I will need all those variables.

assets/js/wcsatt-add-to-cart-variation.js
```diff
@@ -1,42 +1,107 @@
...
+					var allow_one_time_option = false;
+				}
+
+				if ( allow_one_time_option && default_status != 'subscription' ) {
+					var default_subscription_scheme_id = '0';
+				} else {
+					var default_subscription_scheme_id = '1';
+				}
+
+				if ( !allow_one_time_option ) {
```
seb86: Need to set this to true once finished testing.

seb86:
@franticpsyx I have been making progress with the frontend. Still have some areas to finish tweaking but it's looking good. Have you had a chance to look at it yet?

@thenbrent Have you had the chance to look also?

assets/js/wcsatt-add-to-cart-variation.js
seb86: Now set as true.

assets/js/wcsatt-add-to-cart-variation.js
seb86: Turned out I didn't need any of these variables I coped.

includes/class-wcsatt-display.php
seb86: This also allows variations that are not subscriable to work as normal when WooCommerce runs it's own Javascript.

franticpsyx:
Thanks for this, @seb86 :)

We are planning to review your PR this week, and make tweaks/adjustments as needed. Thanks again!

seb86:
@franticpsyx That sounds great. I have a few more Javascript improvements for the frontend that I will be pushing later today. I'm still figuring out how to get the item data added to the cart properly for the variation the customer selected.

There are three bugs that I know exist that I have not got round to fixing yet for the backend.

The first major bug is when adding the options. It stops you from adding more than 3 subscription options and this is because ether the loop or the index ID is not passing correctly and is not returning the correct number each time you add a new option so when you save, the last option overrides all other options you have added which currently labels as option 3.

The second is the ability to re-order the sub options. When trying to drag and drop them, it still thinks I'm trying to drag and drop the whole variation.

And last is a small bug which is suppose to simply do a hide and show on the price fields depending on the price override selected. Upon loading the variations all fields are exposed. Only when changing the price method does it hide the other fields as it should. I must of missed a piece of code when re-writing the Javascript for the variations.

franticpsyx:
> I'm still figuring out how to get the item data added to the cart properly for the variation the customer selected.

@seb86 Ping me if you have any questions. You might need to modify https://github.com/Prospress/woocommerce-subscribe-to-all-the-things/blob/master/includes/class-wcsatt-schemes.php#L175 in order to obtain the saved scheme meta from the variation post id (when the cart item is a variation) instead of the product id. In general once the selected scheme id is posted, the rest should be handled without any modifications apart from this one, as far as I remember.

>There are three bugs that I know exist that I have not got round to fixing yet for the backend.

Let us know when you are done! We'll be working on top of this PR, so it probably makes sense to start reviewing + committing code once you feel confident you are finished.

seb86:
@franticpsyx Thank you for the tip and I will let you know either later today or tomorrow when I will be finished so that you can take over.

seb86:
@franticpsyx It is now more functional at the front end. The discount function will need a little further adjustment. I used a function I found on Stack OverFlow which I have referenced in the Javascript.

One of my test products has a discount that gives me 1.74€ on the product page but on the cart page it shows 1.75€

Also, in issue #30 I mentioned that the sale price of the product level is not being used to calculate a discount. It is doing the same for variable products due to this issue so, for now, the discount calculation in the Javascript is based on the regular price.

With the new front end Javascript, I also made sure that it does not apply for product bundle products so that the scheme data remains the same from the product level. Otherwise, it was adding two subscription levels which is just confusing.

I'm going to try another attempt at the backend but you can start to work on the frontend for now. I have also left console.logs in place until they are ready to be removed.

I have not yet minified the Javascript. Probably be best that you do that with Grunt as I don't have that setup yet.

seb86:
New issue found in the backend. When creating a new variable product the variations have to be saved and then the whole product first in order for the Javascript to allow you to add subscription options. Otherwise the "Add Option" button appears to be dead.

franticpsyx:
@seb86 @thenbrent 

Took it for a spin and had a look at the code in a bit more depth.

**Front-end Feedback**

I like how you are using the `woocommerce-variation-price` div to output the subscription options by replacing the standard price html (the `wcsatt-variation-sub-options` div that you added is redundant, as far as I can tell). 

However, I am not a big fan of using such a large amount of JS to render the options at the client side. I would rather use a single template for rendering the options at the server-side, probably using the `woocommerce_available_variation` filter. I haven't actually implemented that to verify, but it might be possible to make this work without needing to write a dedicated front-end script. The first thing that comes to my mind would be to use the `price_html` variations data field to hold the rendered sub options, and let the core script render that markup without having to write any JS at all. Ideally, I would like to use the same server-side logic/templating to render the sub scheme options for both simple products + variations.

**Admin Feedback**

Using a dedicated "Subscribable" checkbox to enable the Sub functionality is a good UX idea that reduces clutter. I'd rather rename it to something else, but @thenbrent is a native English speaker, so I'd leave that to him.

With that said, I think it's important to port all options found under the original "Subscriptions" tab, especially the ones that control whether the item is still available as a one-off purchase or not, in order to maintain the "Subscribe and Save" functionality. Some of the options could probably use some language/description updates, but I find they are all essential.

Bulk editing is going to be big challenge, too - currently you have added the ability to toggle the "Subscribable" field, but ideally we need a dedicated modal to allow configuring all options found under the original "Subscriptions" tab. For me ( and as already pointed out by @thenbrent ) this is probably going to be one of the biggest challenges, since we should not expect shop managers to add sub options manually under each variation. Having given this some more thought since our discussion with Brent, I'm also not happy with variations support without a complete Bulk editing system. Even for a variable product with 10 variations, entering Sub options manually is going to be a big pain, to the extent of making the feature unusable.

For me the admin work required here is probably 80-90% of the work needed to get this right. If we can't dedicate enough resources to spec and implement this right, I'd rather start with something simpler, such as variable-level support, and move forward with our original vision for StATT, which is to slowly add proper support for more product types.

@seb86 @thenbrent What are your thoughts about the importance of bulk editing here? 

>ideally we need a dedicated modal to allow configuring all options found under the original "Subscriptions" tab

I haven't worked extensively with extending WC at variation-level, so can't estimate how complex this might be off the top of my head.

seb86:
@franticpsyx You are right about the div. I was using it at the time of the early experiments. That can now be removed.

> (the wcsatt-variation-sub-options div that you added is redundant, as far as I can tell).

seb86:
> However, I am not a big fan of using such a large amount of JS to render the options at the client side. I would rather use a single template for rendering the options at the server-side, probably using the woocommerce_available_variation filter.

@franticpsyx I actually already use this filter and the reason why I did it with JS is because variation products use JS in order to display the variation data anyway and without overriding template files I chose to do it this way to be more compatible. Also, the JS only works it's magic if that variation has any scheme data allowing the original WC JS to still work.

seb86:
@franticpsyx @thenbrent My thoughts on the bulk editing.

I admit that having to manually enter every variation can be a pain but we also have to consider on the bulk options we provide. I think it would be silly to set all sub options with the same price would it not?

Also before you can even do a bulk edit for any of the fields in each sub option, they have to exists first.

So you would need a bulk action before doing another such as how many sub options to you want to add to this variation.

Once those sub options have been added can you then do a second bulk action.

So I guess we should also consider if the second bulk actions are part of the variation bulk edit options or provide our own bulk edit option next to the "Add Option" button.

What do you think?

thenbrent:
@seb86 I've had a chance to start looking over this PR now. So far, it adds around 1,500 lines of code to a project which is only around 2,500 line to begin with. The implementation also isn't finished yet, as it would definitely need some form of bulk editing to make it usable, which means it could end up getting close to doubling the size of the existing codebase.

It isn't surprising a lot of code is needed for variation level subscription schemes - that was always my concern with offering variation level support. But I also think much of the code in this PR is duplicating existing code in the extension that could be refactored to be reused instead.

@seb86 try to find ways to implement @franticpsyx suggestions for not duplicating code and reducing the amount of code, including:

* don't replace replacing the standard price html (`wcsatt-variation-sub-options` div)
* use a single template for rendering all front-end subscription options (ideally using `price_html` variations data field to hold the rendered sub options)

In addition, I'd also suggest you:

* refactor the existing subscription options admin/meta box template and JS so that the same template can be used for variations as well as other product types.
* remove the **Subscribable** checkbox - it's not necessary. If Subscription options exist on a product, then that product is subscribable. If they do not, then it is not subscribable. This may also help fix up the related issue where variations that have subscription options do not have a one-off purchase option (which is different to the STATT behaviour of other products).

#### Bulk Editing

For bulk editing, it may be enough to offer the subscription options on the _Subscriptions_ tab on Variable Products, then apply those options to all variations _unless_ some subscription options are set on the variation.

That would allow for variation subscription options without requiring the store manager to set the options on _every_ variation.

It doesn't allow bulk editing of most fields, but if store owners use the variation level subscription schemes, they'll tell us whether they need bulk editing of those fields (like price type, discount amount etc.) and we can add it in with their feedback.

#### Errors

Beyond all of that, I'm also getting a lot of notices logged when adding/saving a variable product with subscription options. Some of those include:

```
[03-May-2016 20:04:40 UTC] PHP Notice:  Trying to get property of non-object in /wp-content/plugins/woocommerce/includes/admin/wc-meta-box-functions.php on line 22
[03-May-2016 20:04:40 UTC] PHP Stack trace:
[03-May-2016 20:04:40 UTC] PHP   1. {main}() /wp-admin/admin-ajax.php:0
[03-May-2016 20:04:40 UTC] PHP   2. do_action() /wp-admin/admin-ajax.php:89
[03-May-2016 20:04:40 UTC] PHP   3. WCS_ATT_Admin::ajax_add_variation_subscription_scheme() /wp-includes/plugin.php:525
[03-May-2016 20:04:40 UTC] PHP   4. do_action() /wp-content/plugins/woocommerce-subscribe-to-all-the-things/includes/admin/class-wcsatt-admin.php:765
[03-May-2016 20:04:40 UTC] PHP   5. WCS_ATT_Admin::variable_subscription_scheme() /wp-includes/plugin.php:525
[03-May-2016 20:04:40 UTC] PHP   6. include() /wp-content/plugins/woocommerce-subscribe-to-all-the-things/includes/admin/class-wcsatt-admin.php:436
[03-May-2016 20:04:40 UTC] PHP   7. do_action() /wp-content/plugins/woocommerce-subscribe-to-all-the-things/includes/admin/views/variable-subscription-scheme.php:29
[03-May-2016 20:04:40 UTC] PHP   8. WCS_ATT_Admin::variable_subscription_scheme_product_content() /wp-includes/plugin.php:525
[03-May-2016 20:04:40 UTC] PHP   9. woocommerce_wp_text_input() /wp-content/plugins/woocommerce-subscribe-to-all-the-things/includes/admin/class-wcsatt-admin.php:696
---
[03-May-2016 20:05:33 UTC] PHP Notice:  Undefined index: subscription_sale_price in /wp-content/plugins/woocommerce-subscribe-to-all-the-things/includes/admin/class-wcsatt-admin.php on line 305
[03-May-2016 20:05:33 UTC] PHP Stack trace:
[03-May-2016 20:05:33 UTC] PHP   1. {main}() /wp-admin/post.php:0
[03-May-2016 20:05:33 UTC] PHP   2. edit_post() /wp-admin/post.php:193
[03-May-2016 20:05:33 UTC] PHP   3. wp_update_post() /wp-admin/includes/post.php:369
[03-May-2016 20:05:33 UTC] PHP   4. wp_insert_post() /wp-includes/post.php:3602
[03-May-2016 20:05:33 UTC] PHP   5. do_action() /wp-includes/post.php:3529
[03-May-2016 20:05:33 UTC] PHP   6. WC_Admin_Meta_Boxes->save_meta_boxes() /wp-includes/plugin.php:525
[03-May-2016 20:05:33 UTC] PHP   7. do_action() /wp-content/plugins/woocommerce/includes/admin/class-wc-admin-meta-boxes.php:221
[03-May-2016 20:05:33 UTC] PHP   8. WCS_ATT_Admin::process_product_meta() /wp-includes/plugin.php:525
---
[03-May-2016 20:05:33 UTC] PHP Notice:  Undefined index: subscription_sale_price in /wp-content/plugins/woocommerce-subscribe-to-all-the-things/includes/admin/class-wcsatt-admin.php on line 306
[03-May-2016 20:05:33 UTC] PHP Stack trace:
[03-May-2016 20:05:33 UTC] PHP   1. {main}() /wp-admin/post.php:0
[03-May-2016 20:05:33 UTC] PHP   2. edit_post() /wp-admin/post.php:193
[03-May-2016 20:05:33 UTC] PHP   3. wp_update_post() /wp-admin/includes/post.php:369
[03-May-2016 20:05:33 UTC] PHP   4. wp_insert_post() /wp-includes/post.php:3602
[03-May-2016 20:05:33 UTC] PHP   5. do_action() /wp-includes/post.php:3529
[03-May-2016 20:05:33 UTC] PHP   6. WC_Admin_Meta_Boxes->save_meta_boxes() /wp-includes/plugin.php:525
[03-May-2016 20:05:33 UTC] PHP   7. do_action() /wp-content/plugins/woocommerce/includes/admin/class-wc-admin-meta-boxes.php:221
[03-May-2016 20:05:33 UTC] PHP   8. WCS_ATT_Admin::process_product_meta() /wp-includes/plugin.php:525
---
[03-May-2016 20:05:33 UTC] PHP Notice:  Undefined index: subscription_period_interval in /wp-content/plugins/woocommerce-subscribe-to-all-the-things/includes/admin/class-wcsatt-admin.php on line 343
[03-May-2016 20:05:33 UTC] PHP Stack trace:
[03-May-2016 20:05:33 UTC] PHP   1. {main}() /wp-admin/post.php:0
[03-May-2016 20:05:33 UTC] PHP   2. edit_post() /wp-admin/post.php:193
[03-May-2016 20:05:33 UTC] PHP   3. wp_update_post() /wp-admin/includes/post.php:369
[03-May-2016 20:05:33 UTC] PHP   4. wp_insert_post() /wp-includes/post.php:3602
[03-May-2016 20:05:33 UTC] PHP   5. do_action() /wp-includes/post.php:3529
[03-May-2016 20:05:33 UTC] PHP   6. WC_Admin_Meta_Boxes->save_meta_boxes() /wp-includes/plugin.php:525
[03-May-2016 20:05:33 UTC] PHP   7. do_action() /wp-content/plugins/woocommerce/includes/admin/class-wc-admin-meta-boxes.php:221
[03-May-2016 20:05:33 UTC] PHP   8. WCS_ATT_Admin::process_product_meta() /wp-includes/plugin.php:525
---
[03-May-2016 20:05:33 UTC] PHP Notice:  Undefined index: subscription_period in /wp-content/plugins/woocommerce-subscribe-to-all-the-things/includes/admin/class-wcsatt-admin.php on line 343
[03-May-2016 20:05:33 UTC] PHP Stack trace:
[03-May-2016 20:05:33 UTC] PHP   1. {main}() /wp-admin/post.php:0
[03-May-2016 20:05:33 UTC] PHP   2. edit_post() /wp-admin/post.php:193
[03-May-2016 20:05:33 UTC] PHP   3. wp_update_post() /wp-admin/includes/post.php:369
[03-May-2016 20:05:33 UTC] PHP   4. wp_insert_post() /wp-includes/post.php:3602
[03-May-2016 20:05:33 UTC] PHP   5. do_action() /wp-includes/post.php:3529
[03-May-2016 20:05:33 UTC] PHP   6. WC_Admin_Meta_Boxes->save_meta_boxes() /wp-includes/plugin.php:525
[03-May-2016 20:05:33 UTC] PHP   7. do_action() /wp-content/plugins/woocommerce/includes/admin/class-wc-admin-meta-boxes.php:221
[03-May-2016 20:05:33 UTC] PHP   8. WCS_ATT_Admin::process_product_meta() /wp-includes/plugin.php:525
---
[03-May-2016 20:05:33 UTC] PHP Notice:  Undefined index: subscription_length in /wp-content/plugins/woocommerce-subscribe-to-all-the-things/includes/admin/class-wcsatt-admin.php on line 343
[03-May-2016 20:05:33 UTC] PHP Stack trace:
[03-May-2016 20:05:33 UTC] PHP   1. {main}() /wp-admin/post.php:0
[03-May-2016 20:05:33 UTC] PHP   2. edit_post() /wp-admin/post.php:193
[03-May-2016 20:05:33 UTC] PHP   3. wp_update_post() /wp-admin/includes/post.php:369
[03-May-2016 20:05:33 UTC] PHP   4. wp_insert_post() /wp-includes/post.php:3602
[03-May-2016 20:05:33 UTC] PHP   5. do_action() /wp-includes/post.php:3529
[03-May-2016 20:05:33 UTC] PHP   6. WC_Admin_Meta_Boxes->save_meta_boxes() /wp-includes/plugin.php:525
[03-May-2016 20:05:33 UTC] PHP   7. do_action() /wp-content/plugins/woocommerce/includes/admin/class-wc-admin-meta-boxes.php:221
[03-May-2016 20:05:33 UTC] PHP   8. WCS_ATT_Admin::process_product_meta() /wp-includes/plugin.php:525
```

#### Kiss of Death

It's worth mentioning that the type of extra maintenance overhead this PR introduces is not something I'm inclined to merge and take responsibility for at the moment. We'll see how you go @seb86 with reducing the code duplication, but if we can't get it significantly smaller PR to implement this feature, I think it's going to be better handled in a separate plugin for your client rather than in STATT core itself.

It can be the kiss of death for free plugins to include more code than can be feasibly maintained without a commercial model (which we don't have for STATT yet). I'd hate to do that to an extension with so much potential as STATT.

franticpsyx:
@thenbrent I think I agree on all points.

I think @seb86 deserves a lot of credit for the work done here, but I'm also skeptical about merging so much new code without having agreed at least on a basic spec. @thenbrent 's comment above provides a rough "spec" that I'd like to see implemented in a new branch.

>For bulk editing, it may be enough to offer the subscription options on the Subscriptions tab on Variable Products, then apply those options to all variations unless some subscription options are set on the variation.

This is a spot-on observation which gives us another reason to start with Variable-level support, then push forward with other tasks and leave Variation-level for later, when there is more measurable demand and feedback. Variable-level options will definitely mitigate the lack of a proper Bulk-editing system later on, when we deliver the feature at Variation-level, since the Variation-level options could be added when there's a need to "override" the product-level definitions.

Ideally, the front-end implementation should be future-proof enough to require only minimal additions (if any at all) once variation-level options are added. This will allow @seb86 to migrate his existing work on top of the new branch, or simply deliver his current work and rest assured that the variation-level post meta will have the same structure as the one we currently use for Simple products.

I'm happy to start working on this if this all sounds good. Rough ETA - we should have a Variable-level support draft to review next week.

seb86:
@franticpsyx So do you think it would be best then to put my code in a new plugin as an extension to this extension for my client until this is dealt with properly?

franticpsyx:
>So do you think it would be best then to put my code in a new plugin as an extension to this extension for my client until this is dealt with properly?

If you think this is possible then, yes (might be a bit hard without adding a generous amount of hooks).

Alternatively, you might be able to deliver your fork and update it with the `master` branch version when variation-level support gets added, since I don't think we'll change the metadata structure for variations.

thenbrent:
> I think @seb86 deserves a lot of credit for the work done here

Yes, I should have acknowledged that. Clearly a lot of work has gone into this, all props for that hard work. I hope we can make sure it is a valuable addition for your client and hopefully others who need it in future if we can make it work.

> without adding a generous amount of hooks

More than happy to add any hooks required to get this into a separate extension/plugin to begin with.

> the front-end implementation should be future-proof enough to require only minimal additions (if any at all) once variation-level options are added

Yes, that would be perfect, if possible.

> I'm happy to start working on this if this all sounds good. Rough ETA - we should have a Variable-level support draft to review next week.

Sounds good to me.

franticpsyx:
Closing as we are starting with Variable-level support on a separate branch.

Issue #2, opened 2015-08-18
Variable products support


franticpsyx:
Currently we do not allow subscription schemes to be defined for variable products -- when we do it, we could make this work at variation level, or we could simply ignore variations and use the existing "Subscriptions" tab to define subscription schemes at parent level.

louis2b:
I was thinking at this as I'd find it very useful to be able to subscribe to different skus. 
It may be obvious but I was thinking, how it could look in terms of ux, for variable products. 

I see 2 options to be able to define subs schemes at variation level
1. Having subscription fields appear in the different variations settings (in the variations tab)
2. keeping a subs tab, and for each subs scheme be able to give the variations to which it should be applied 

Option 2 seems much more interesting as there could be tons of variations and setting subs schemes 1by1 could be frustrating. 

Plus I agree with @franticpsyx in #1 : variations is sthg that defines the products, and subscriptions schemes is more how you sell it. So they should be kept separated.

franticpsyx:
Added to the 1.1 milestone - cc @thenbrent

Ahrengot:
Is there a rough time frame for when 1.1 will be ready?

franticpsyx:
>Is there a rough time frame for when 1.1 will be ready?

Hard to say - however, I can definitely say that we will be able to allocate more resources to SATT in Q2.

Ahrengot:
Alright I'm still a little unclear on the time frame. I'm trying to figure out if we're talking about weeks or months, but it sounds like months — Correct?

franticpsyx:
@thenbrent Can we perhaps brainstorm a few ideas about this to see what we want/need to implement?

@Ahrengot The main issue here is that we haven't specced this feature at all, so it's still unclear what the workload is going to be. We don't have a roadmap in place for SATT development, but we do have a pretty good idea of the features we want to prioritize - this is one of them. It's still hard to give an ETA, though - sorry I can't be more specific but the open source nature of the project makes it difficult to make any commitments.

seb86:
@franticpsyx and @Ahrengot I think option one would be better for adding subscriptions to variation products.

I think it would be confusing to some users to go between tabs in order to apply subscription options per variation.

Plus there is the issue of the data being saved. If you placed the subscription options under it's own section, would it not require more work to identify each subscription option per variable option.

thenbrent:
> Can we perhaps brainstorm a few ideas about this to see what we want/need to implement?

[...]

PR #35, opened 2016-05-14
Variable products support

franticpsyx:
@thenbrent @seb86 

A draft for variable products support at product-level. Some things worth noting:

- Feel free to ignore the individual commits and look at the changes directly. The "real" additions/changes are minimal, if you ignore some restructuring and obsessive-compulsive tweaks to existing code and inline docs.
- In the admin area, I have left both "price override" methods active (Discount + Override), although in #2 we discussed that the **Override** option doesn't seem too meaningful/relevant for variable products when adding subscription options at product-level. I figured, however, that it might be useful if the intent is to override the price of **all variations** and offer them at the same discounted price when subscribing to one of them. You will also notice that the admin "Subscriptions" tab option descriptions for variable products are slightly modified in some cases - for instance the "Override product" Subscription Price option has been renamed to "Override all variations" and the "Inherit from product" option has been renamed to "Inherit from chosen variation".
- All front-end work associated with displaying subscription options is contained in https://github.com/Prospress/woocommerce-subscribe-all-the-things/commit/1230c7016916905182f42697a48bc7dc825220b1 . Note that the task of updating the options is handled by the core variations script, since the options markup is replacing the `price_html` field in the `variations_data` json object. The markup is rendered using the same logic + template that renders the options in the Simple product case.
- I got a bit stuck with modifying the price string returned by the `get_price_html` method, but I think my modifications to https://github.com/Prospress/woocommerce-subscribe-all-the-things/blob/issue-2/includes/class-wcsatt-display.php#L428 should be enough. 
- While variable products support at such a basic level has been mostly painless, I have had a chance to roughly estimate the difficulty for implementing this at variation-level. Ironically, one of the toughest problems to tackle when/if we approach this at variation level is going to be the price string "calculation" (filtering the output of `get_price_html`), since each variation will have its own set of sub schemes. When the time comes, we'll see how to handle this "challenge", but let's just note it down for now.

Let me know what you find :)


seb86:
@franticpsyx Code looks OK to me just quickly reading through it. I will test it out later today.

franticpsyx:
@thenbrent @seb86 Merging, then! If you happen to find any bug, feel free to open a separate issue. For now, let's consider #2 closed until we decide to take this to variation level.

Issue #20, opened 2016-02-03
Price overrides compatibility with per-item priced Bundles/Composites/MnM


franticpsyx:
ref: https://github.com/Prospress/woocommerce-subscribe-to-all-the-things/pull/13#issuecomment-176881998

franticpsyx:
Added to the 1.1 milestone.

franticpsyx:
Task list:

- Integrate with Bundles/Composites front-end app to obtain up-to-date price-overridden totals and display them next to each SATT sub option.
- Ensure that when a single SATT scheme exists and the "Force Sub" option is checked, all product prices will be overridden on the server side before loading the page. In the case of CP, some additional work is required to ensure that prices in ajax `show_composited_product` requests are filtered.
- Ensure price html strings are correct.


To make all this easier, all price-related functions will be ported into a new class whose purpose will be to handle price overrides / discounts. Some work on the CP / PB side will also be needed to ensure that the JS apps can return totals based on custom price inputs.

PR #41, opened 2016-06-03
Issue #20: Price overrides compatibility with per-item priced Bundles/Composites

franticpsyx:
Creating this to keep track of progress in relation to #20 

New code includes:

- A front-end script to integrate with the front-end Bundles and Composites scripts - this has the purpose of updating the totals displayed next to SATT options based on parameters added to the options markup in a `data` attribute.
- Some hooks in `class-wcsatt-integrations.php`, which mainly help with the case of having 1 satt scheme with the force-subscription option enabled. In this case, all price discounts / overrides should be applied on the server side when loading the product, since we have no options to show to the user. When multiple options exist, the displayed prices and totals calculations (apart from the html price string shown in the product summary) are displayed / carried out as usual. However, if there are SATT options with price overrides, we are including dynamically updated price-html totals based on the associated discounts.

The rest is mainly refactoring of the price overriding logic to allow more complex integrations. Most of the logic is now located in a separate class, `class-wcsatt-scheme-prices.php`.

There is still a lot of testing to do and some pending work in relation to CP support - more specifically, when loading Component Options details via ajax and the Composite only has a single SATT scheme with the "Force Sub" option enabled, we need to ensure that prices in the returned data have been filtered according to the settings of the sub scheme attached on the Composite.

Additionally, I have found some minor issues with Add-ons attached on composited/bundled products.


franticpsyx:
@seb86 Thank you, Seb!

Note this won't work until new versions for both CP and PB are available!

Issue #1, opened 2015-08-18
Additional subscription scheme settings


franticpsyx:
In its current state, the mini-extension is missing the following sub settings from the product-level subscription scheme UI:
- trial length
- trial period
- sign up fee

It's also missing the "Limit Subscription" setting, and the "Synchronise Renewals" setting.

These are obviously important, but since we rely on Subs core for all associated functionality, adding these in should be a trivial matter once everything else gets reviewed.

felipesantana:
I might have some issue with the this plugin, there is no "Subscription Option" on the tab "Subscription", I clicked "add option" and nothing happens. Did I miss some configuration? Heres a screenshot and thanks for this plugins, if I can get this working it will save my life ;D

Screenshot: http://cl.ly/image/0e0M1n3F3F1L

franticpsyx:
@felipesantana The issue you are reporting is irrelevant to this thread - please open another one. Also I'm not able to replicate this issue. Please check your browser console for errors, and post your findings in a new ticket. Thanks!

franticpsyx:
Submitted ideas:

- add the possibility for the store manager to set a % discount if the user subscribes to a product
- add sync renewal options

thenbrent:
> add the possibility for the store manager to set a % discount if the user subscribes to a product

Cool, like Amazon's "Subscribe and Save" system (though I think they've trademarked that phrase so we'll need to call it something else).

> add sync renewal options

Hopefully we can most of the exiting methods in `WC_Subscriptions_Synchroniser` to handle this. At very least for the logic to handle everything after the products are made into cart items. To actually sync the subscription options for a product might need a bit more custom code. Although ideally we also use `WC_Subscriptions_Synchroniser::subscription_product_fields()` to display the sync fields (it will need extra JS to link the sync period with that subscription options chosen billing period though).

franticpsyx:
>Cool, like Amazon's "Subscribe and Save" system (though I think they've trademarked that phrase so we'll need to call it something else).

This should already work in the https://github.com/Prospress/woocommerce-subscribe-to-all-the-things/tree/issue-13-patches branch - there are some issues with per-item-priced bundle types and the way totals show up in the product details page, but other than that, I think we are good.

Will run some more tests when I get back from Cape Town ;)

franticpsyx:
@thenbrent Note we've allowed both the ability to set a % discount **and** the ability to override the Regular/Sale prices at subscription-scheme level.

[...]

Issue #42, opened 2016-06-09
SATT Products API


franticpsyx:
### Objective

Currently, most logic and data needed for "converting" `WC_Product` objects to subscription-enabled objects are scattered all over the place, for instance:
1. scheme data is only accessible from post meta, while 
2. the active subscription scheme id is stored only in the cart item array
3. subscription details are stored as object properties by WCS

(1) and (2), as well as all associated logic for "converting" a product to a subscription in the cart is currently implemented using static functions, simply organized in suitably named classes.

Obviously, we need a more formal, object-oriented way to: 
- manage the subscription schemes available with a WC product, 
- access its subscription status/details (regardless of its type) and 
- convert its subscription state,

not only in cart context, but in a way that's more tightly integrated with the `WC_Product` class.
### Ideal Case

Ideally, if this would all be part of WC core, the Abstract Product class would include properties and methods to handle the "Subscription Nature" of a product internally, and set any properties associated with it (billing periods, intervals, trial, prices, sign up prices, etc) when converting from/to a non-recurring billing state to a recurring one.

This would allow a product to freely switch between states, while existing methods could ideally take this into account and modify their output, for instance:
- `get_price` and other price getters could modify the returned value depending on the price(s) associated with the active scheme, if set, otherwise return the standard non-recurring (one-off) price.
- `get_price_html` could try to generate an output based on the active subscription (if set), otherwise return something meaningful based on a stored `min_price_subscription_scheme`.

From a cart perspective, this core-embedded handling would allow WC to set the active subscription scheme with the item session data, and "convert" it when instantiating the product class, similar to how products are instantiated using their `product_id` alone - only we'd have an additional `subscription_scheme_id` to pass to the constructor.

From an order perspective, `get_product_from_item` would use a similar method to instantiate the object using a `subscription_scheme_id` field stored in the line item.

This would allow products to have a _dual_ nature - the Subscription and the Product, distinct from one another but unified in the same object.

(Do we agree with this as a utopian solution to start with?)
### Possible Solutions

What's the closest we can get to our Utopia, if we agree we accept it as such?
- Somehow dynamically add properties and methods to all instantiated products on init. If PHP was JS, we would have likely intervened with the Abstract Product prototype, or appended "methods" and "properties" at some point during run time. I don't think we have this option.
[...]

PR #47, opened 2016-06-21
Additional subscription scheme settings

seb86:
This push is in reference to issue #1.

Changes made are simply adding filters to allow the full use of Subscriptions within this extension and allow other extensions to connect with such as [this one](https://github.com/seb86/woocommerce-subscribe-to-all-the-things-signup-trial-add-on).

Also did a version number change.


franticpsyx:
@seb86 Green light from me - I'll give you some more time as we think about #42 just in case you need to add anything else in v1.1.

seb86:
@franticpsyx Thank you.

includes/class-wcsatt-display.php
```diff
@@ -526,6 +536,43 @@ public static function filter_price_html( $price, $product ) {
...
+
+		$product_type = $product->product_type;
+
+		$product_schemes = get_post_meta( $product->id, '_wcsatt_schemes', true );
+
+		$force_subscription = get_post_meta( $product->id, '_wcsatt_force_subscription', true );
+
+		if ( in_array( $product_type, WCS_ATT()->get_supported_product_types() ) && $product_schemes ) {
+
+			if ( $force_subscription == 'yes' ) {
```
thenbrent: This should be a yoda condition, i.e. `'yes' == $force_subscription`, to conform to the [WP coding standards](https://make.wordpress.org/core/handbook/best-practices/coding-standards/php/#yoda-conditions).

thenbrent:
@seb86 LGTM. One issue with use of yoda condition but otherwise 👍 

I'd also suggest not updating version numbers in PRs. That can be done at deployment time rather than during development.

includes/class-wcsatt-display.php
seb86: @thenbrent I obviously still have much to learn. I did not know of this Yoda condition. I will push a quick fix.

seb86:
@thenbrent I changed the version during development of the extension I'm building. I will remember next time to put it back to the original version before committing.

seb86:
@franticpsyx @thenbrent Any chance this will be merged soon?

includes/class-wcsatt-display.php
```diff
@@ -526,6 +536,43 @@ public static function filter_price_html( $price, $product ) {
...
+			if ( 'yes' == $force_subscription ) {
+
+				if ( $product->variation_id > 0 ) {
+
+					$button_text = __( 'Sign Up this Variation', WCS_ATT::TEXT_DOMAIN );
+
+				} else {
+
+					$button_text = get_option( WC_Subscriptions_Admin::$option_prefix . '_add_to_cart_button_text', __( 'Sign Up Now', WCS_ATT
+::TEXT_DOMAIN ) );
```
thenbrent: This doesn't need to be on a new line. 😄

includes/class-wcsatt-display.php
```diff
@@ -526,6 +536,43 @@ public static function filter_price_html( $price, $product ) {
...
+
+		$force_subscription = get_post_meta( $product->id, '_wcsatt_force_subscription', true );
+
+		if ( in_array( $product_type, WCS_ATT()->get_supported_product_types() ) && $product_schemes ) {
+
+			if ( 'yes' == $force_subscription ) {
+
+				if ( $product->variation_id > 0 ) {
+
+					$button_text = __( 'Sign Up this Variation', WCS_ATT::TEXT_DOMAIN );
```
thenbrent: This button text feels a little too custom to me to include in core. I'd probably instead always use the value of `WC_Subscriptions_Admin::$option_prefix . '_add_to_cart_button_text'` and if the store owner wants custom text for variations, they can use the `'wcsatt_add_to_cart_text'` filter to set that.

thenbrent:
@seb86 sorry for the delay. I've left two small comments, otherwise this looks good to me and can be merged once those are fixed up and @franticpsyx gives it the 👍

franticpsyx:
@seb86 Took a look and basically agree with @thenbrent 's comments. I'll merge and release v1.1 when the points Brent raised are fixed.

I hoped to have some issues reported for v1.1 RC1 before calling v1.1 final - also waited for a last word from you to merge but caught up by work in the meanwhile.

seb86:
OK @thenbrent and @franticpsyx. I will correct this today and get back to you both.

seb86:
Done. Any chance issue #49 could be addressed quickly @franticpsyx before the release?

includes/admin/class-wcsatt-admin.php
```diff
@@ -36,7 +36,7 @@ public static function init() {
 		add_action( 'woocommerce_product_write_panels', __CLASS__ . '::product_write_panel' );
 
 		// Processes and saves the necessary post meta.
-		add_action( 'woocommerce_process_product_meta', __CLASS__ . '::process_product_meta' );
+		add_action( 'woocommerce_process_product_meta', __CLASS__ . '::process_product_meta', 15, 1 );
```
seb86: I also made sure it saves the meta data after WooCommerce has done it's first, not before.

franticpsyx:
@seb86 

I have a feeling that 6bf86da shouldn't be in SATT, but I might be wrong. I'm not familiar with the Sub Shortcodes extension but this piece of code doesn't look like it's making SATT do something for its self that it wouldn't do otherwise. If it's adding SATT support into another plugin, then it should go in there (conditionally on the presence of SATT).

>Any chance issue #49 could be addressed quickly

Yep, will have a look once we decide what to do with 6bf86da and finally merge this.

franticpsyx:
@seb86 please avoid adding irrelevant commits to this PR to keep things focused - 342cf14 is not relevant with "Additional subscription scheme settings" or adding hooks, which was the intent of this PR :)

seb86:
@franticpsyx Should I then remove commit 6bf86da so you can merge. You may be right about adding it to the subscription shortcodes extension instead.

franticpsyx:
Thanks @seb86 - yep, if you can remove 6bf86da I'll have a final look and then merge 👍

seb86:
@franticpsyx Great. :+1:

seb86:
Ready, Set, Merge, Release  :package:

franticpsyx:
@seb86 Tweaked a few things in https://github.com/Prospress/woocommerce-subscribe-all-the-things/commit/07765c19fa5f23346947ed0c3db52370f1f05a21 

To be honest, I don't think this filter was needed - https://github.com/Prospress/woocommerce-subscribe-all-the-things/commit/07765c19fa5f23346947ed0c3db52370f1f05a21#diff-af820c6be360bb1a8e112cfab4d897ddR167

There's already a filter that allows you to change the option description a few lines below. I'd be happy to remove the `wcsatt_single_product_subscription_scheme_price_html` filter if you can use the one that follows.

seb86:
@franticpsyx I will be able to go over it tomorrow.

seb86:
> To be honest, I don't think this filter was needed - 07765c1#diff-af820c6be360bb1a8e112cfab4d897ddR167

@franticpsyx I'm pretty sure I put that filter there so that I could return the sign up fee and trial. So now I need to change the filter in my extension which I will do tomorrow.

franticpsyx:
>I put that filter there so that I could return the sign up fee and trial. So now I need to change the filter in my extension which I will do tomorrow.

@seb86 Can you use this one instead? - https://github.com/Prospress/woocommerce-subscribe-all-the-things/blob/master/includes/class-wcsatt-display.php#L181

I totally missed there was a filter when looking at your diff - perhaps you missed that as well?

seb86:
> @seb86 Can you use this one instead?

@franticpsyx No, because it needs to filter the function `WC_Subscriptions_Product::get_price_string` on [line 167](https://github.com/Prospress/woocommerce-subscribe-all-the-things/blob/master/includes/class-wcsatt-display.php#L167).

seb86:
@franticpsyx I have [changed the filters](https://github.com/seb86/woocommerce-subscribe-to-all-the-things-signup-trial-add-on/commit/402a3db5531672c3bc7f312a0ff38e2af027c3f3) used in my extension to match your tweak.

Issue #53, opened 2016-07-25
Support One Time Shipping


seb86:
How comes this is not yet available and what is needed to support it?

I assume we would put it in the same place as the default subscription product types under "Shipping" as it would apply to all subscription schemes.

![One Time Shipping](https://cldup.com/EUke3MoDiR.png)

To me, I see the schemes are only for the pricing and billing conditions.

@thenbrent @franticpsyx Do you both agree?

seb86:
Here is my pull request. https://github.com/Prospress/woocommerce-subscribe-all-the-things/pull/54

franticpsyx:
@seb86 The long-term goal is to have all subscription product-type options available in every subscription scheme. We'll be pushing for this once we have dealt with #42 

I'd like to clean up the scheme logic before building new features into it, but you are right that this is one of the things that needs to be looked into.

seb86:
@franticpsyx I think what I have done for now works great as a temporary measure and one that is needed now even if we don't add any other features for a while until you have refactored how the product subscription scheme data is saved. What do you think?

franticpsyx:
@seb86 It's really a matter of prioritization - no matter how big or small, I'd rather avoid adding a feature until we have a clear idea of what API changes need to go into v1.2 or v2.

seb86:
@franticpsyx OK that's fine. I just did it quickly because a client is using this extension for composite products and emailed me to say the standard subscription option for shipping one time was not there which I thought it was already.

PR #54, opened 2016-07-25
One Time Shipping

seb86:
One Time Shipping appears to work out of the box.

I also corrected an undefined price variable for the dummy product.


includes/admin/class-wcsatt-admin.php
```diff
@@ -35,6 +35,9 @@ public static function init() {
 		// Creates the panel for configuring subscription options.
 		add_action( 'woocommerce_product_write_panels', __CLASS__ . '::product_write_panel' );
 
+		// Add subscription shipping options on edit product page.
+		add_action( 'woocommerce_product_options_shipping', __CLASS__ . '::subscription_shipping_fields' );
```
thenbrent: Given the `WCS_ATT_Admin::subscription_shipping_fields()` function outputs markup identical to `WC_Subscriptions_Admin::subscription_shipping_fields()`, you might as well just use the later and save the duplication. You'll probably need to make sure it will never lead to `WC_Subscriptions_Admin::subscription_shipping_fields()` being called when it's undefined (but hopefully SATT isn't being loaded/executed when Subscriptions isn't defined).

thenbrent:
@seb86 this looks pretty good and I'm happy to merge with a few small changes (keeping the comments on https://github.com/Prospress/woocommerce-subscribe-all-the-things/issues/53 about scope/future direction in mind).

In addition to the above comment about using `WC_Subscriptions_Admin::subscription_shipping_fields()`, one other thing you'll probably need to do is make sure the One Time Shipping fields are hidden when no subscription options are defined for a product. It doesn't make sense to display it for non-subscription products (they are one time shipping by default). Also be careful to make sure that show/hide logic doesn't conflict with core subscription products (i.e. you don't want to hide the One Time Shipping fields on a subscription product just because it has no subscription schemes defined).

seb86:
@thenbrent I hope this is better. I also made a few corrections to remove a few bugs I found.

seb86:
Can this now be merged?

thenbrent:
Hey @seb86, I'm trying to test this now, but I can't pull down your patches. Did you delete your fork? I need access to `seb86:master`.

One other concern is that we had to introduce some changes to show/hide the One Time Shipping toggle based on whether a subscription product had a free trial or was syncrhonised, for details see https://github.com/Prospress/woocommerce-subscriptions/pull/1544

That won't be an issue with SATT out-of-the-box, but it's something that you might need to consider with https://github.com/seb86/woocommerce-subscribe-to-all-the-things-signup-trial-add-on

seb86:
@thenbrent Weird. I haven't removed my fork and for some reason, I can't add you as a collaborator to my fork because it says you are not a GitHub user.

seb86:
OK it found you this time. You are now added to the fork.

seb86:
> That won't be an issue with SATT out-of-the-box, but it's something that you might need to consider with https://github.com/seb86/woocommerce-subscribe-to-all-the-things-signup-trial-add-on

I will add support in my extension once you get this merged first.

thenbrent:
Ah OK I see it's at https://github.com/seb86/woocommerce-subscribe-to-all-the-things now.

I can't fetch or pull from that repo, I get the error:

```
fatal: 'clean' does not appear to be a git repository
fatal: Could not read from remote repository.

Please make sure you have the correct access rights
and the repository exists.
```

> You are now added to the fork.

I shouldn't need to be added to the fork, I just couldn't find the fork as it has the old repo name/URL. Can you try deleting me from the fork and I'll see if GitHub will let me access it?

seb86:
Oh you changed the slug of the repo. I see. OK, I will remove you from the fork. Give me a sec.

seb86:
OK. Try again.

thenbrent:
Working now. It was actually an error on my part by the looks. Testing now...

thenbrent:
Test results:
- [x] Shipping options displayed only for initial order on cart/checkout: https://cloudup.com/iEcrnm1Bbn1
- [x] Shipping charged correctly for initial order: https://cloudup.com/ccsc4H56ivm
- [x] Shipping _not_ added as a line item on Subscription: https://cloudup.com/cXcjLJ37G0l
- [x] Shipping _not_ charged for renewal orders: https://cloudup.com/cOQ1lcNmvjo

LGTM. Nice addition @seb86, thanks! 👍

seb86:
Your welcome @thenbrent

PR #65, opened 2016-09-09
Tidied the subscriptions product data layout and fixed an issue with show/hide schemes

seb86:
This makes the admin UI a little cleaner and fixed an issue.

### DIFF
--- a/assets/css/wcsatt-write-panels.css
+++ b/assets/css/wcsatt-write-panels.css
@@ -74,16 +74,18 @@
 	margin-top: 10px;
 	display: block;
 	float: left;
+	clear: left;
 }
 
 #wcsatt_data .subscription_scheme_product_data label {
-	margin: 0 0 0 -220px;
-	width: 200px;
+	margin: 0 14px 0 0;
+	width: auto;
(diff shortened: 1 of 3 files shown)


seb86:
Synchronize field on the subscriptions tab

The fields are now aligned and now allows the use of grouping fields properly. For example, the extensions I have written can be placed in their own group which is displayed nicely per row without interfering with SATT's fields.

franticpsyx:
The Discount & other fields currently added by SATT under the Subscription Price field are basically grouped "under" it -- this is why they are aligned right, leaving the Subscription Price label aligned left.

If you need to add additional fields you can simply add another `form-field` element at a later `subscription_scheme_content` priority.

I don't see a reason to wrap the existing options in an `options_group` element since all scheme options are few and tightly related anyway - they are all subscription scheme options.

For now, let's have all sub scheme options follow the same pattern of left label / right field alignment. If later on we need to introduce some sort of grouping, I'd rather look at the options we need to group and possibly arrange them in tabbed sections to make things cleaner, than just display all options in a long list.

PR #80, opened 2016-11-23
Create composer.json

philschoefer:
I want to suggest to add a composer.json file so that this extension can be integrated into a composer package easily. Is it too early for that?

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -1,6 +1,6 @@
 {
   "name": "prospress/woocommerce-subscribe-all-the-things",
-  "description": "The goal of the extension is to provide subscription support for non-subscription product types, allowing you to offer any product in your store as a subscription product.",
+  "description": "The goal of this experimental extension is to provide subscription support for non-subscription product types, allowing you to offer any product in your store as a subscription product. Don't use for production.",
   "homepage": "https://github.com/prospress/woocommerce-subscribe-all-the-things",
   "type": "wordpress-plugin",
   "license": "GPL-3.0",


franticpsyx:
I'd give this the green light but this is a feature plugin in non-stable state, which means that it might not be very wise to use it as a dependency in the long run.

@thenbrent thoughts?

thenbrent:
Fine by me to add it. The `description` should be more upfront about the caveats of the extension though. The GitHub repo's description would be a good option, i.e.

> Experimental extension for linking WooCommerce Subscriptions with product types created by other extensions, like Composites and Bundles. Beta status - use at your own risk! — Edit

Its [`homepage` is also not sellwithwp.com](https://github.com/Prospress/woocommerce-subscribe-all-the-things/pull/80/files#diff-b5d0ee8c97c7abd7e3fa29b9a27d1780R4). This GitHub repo's URL should be used for its `homepage` instead.

philschoefer:
Thank you @franticpsyx and @thenbrent. I will integrate the changes you're mentioning.

The thing for me is that I'm building a new project right now that will need the functionality your new plugin is providing. The client has requested that feature so for me its between choosing to build the functionality myself and from scratch or supporting what you've already started.

@thenbrent When you mention the caveats, what are those in specific at the moment? Are there specific bugs (other than what's already mentioned in the "issues" section) that I should be aware of if I want to use this as a basis for my new project?

I'd rather help make this bug free than starting from scratch. Or is it really really far away from being production ready and that's not a good plan altogether?

philschoefer:
Sorry it took me so long to do this. Composer.json is now updated. Thank you guys!

franticpsyx:
@thenbrent Leaving this to you as I have very limited experience with composer. Only use it to install a few things such as phpunit and little else.

thenbrent:
LGTM. Thanks @philschoefer, and sorry for the delay in merging.

philschoefer:
No Problem, thanks guys!

Issue #102, opened 2017-03-15
SATT product in renewal after failed order is not converted to sub


growdev:
To reproduce:

- WC 2.6.14
- WCS 2.0.20
- STATT 1.1.2
- Add a subscription product to cart and checkout
- Our site is using Auth.net so I changed the Billing Zip code to `46282` per [this](https://developer.authorize.net/hello_world/testing_guide/)
- Force a renewal of the subscription
- The order fails and the subscription is placed On Hold
- Login as the customer and go to the `/my-account/subscriptions` page
- Select the On Hold subscription
- Press "Pay" next to the subscriptions failed renewal order
- On the checkout page notice that the subscription scheme is not added to the item

h/t to @javorszky for a similar fix in #68 

PR to follow

javorszky:
The downside of this fix is that if the product's subscription option is discounted, it will be doubly discounted.

If the original price is $15, there's a 10% discount, the initial checkout will be $13.50.

If we're paying for a failed renewal order, that 10% discount gets applied to the $13.50, and it becomes $12.15, which is not ideal.

Any help would be appreciated @franticpsyx, if you have the time / ideas.

franticpsyx:
@growdev @javorszky -- thanks for the report and feedback folks!

We are currently working on #42 and #96 which will introduce some architecture changes and WC 3.0 compatibility (could justify a bump to 2.0). I can definitely have a look at this as part of that refactor.

javorszky:
Awesome, thank you :). I'll read those issues a wee bit later, they are loong. :D

franticpsyx:
You might want to take a look at the `issue-42` branch directly. There's a lot of useful inline commentary there -- the commits should also allow you to understand the objective of the refactor.

I will open a PR to discuss the changes when I'm done refactoring the entire (small) codebase to utilize the new Product API.

growdev:
Thanks for the info @franticpsyx !

franticpsyx:
Merged the work done so far on #96 and #42 in the `2.0` branch. There are a number of issues that are being worked at the same time since the entire codebase is being re-written. Isolating each issue in a separate branch is impractical, but it should be fairly straightforward to follow what's happening - see https://github.com/Prospress/woocommerce-subscribe-all-the-things/compare/master...2.0

franticpsyx:
This _should_ be resolved in `2.0`. Feel free to take that branch for a spin and post feedback!

javorszky:
[...]

PR #103, opened 2017-03-15
Fix for #102, paying for failed renewal order with subscription item

growdev:


franticpsyx:
See #102 -- this _should_ be fixed in `2.0`. If you can take a look and confirm, I'd be eternally grateful! Closing as it's a bad time to merge anything other than`2.0` into master.

javorszky:
At WCEU, I'll try to use some of the breaks to give it a go.

javorszky:
Heya! I gave it a go. It works with resubscribe, the cart contents are correct for paying for failed product, but I couldn't check out due to Stripe handling broken tokens differently (previously it would have failed, now it's an error).

@growdev if you have time, can you double check it on the client's site? I think it'll work.

PR #135, opened 2017-06-15
Fixed overlapping labels in Admin > Subscription Options

rezzz-dev:
There were overlapping labels of form fields within the panel for the Interval section

### DIFF
--- a/assets/css/wcsatt-write-panels.css
+++ b/assets/css/wcsatt-write-panels.css
@@ -115,6 +115,10 @@
 	float: left;
 }
 
+#wcsatt_data .subscription_scheme_data p._satt_subscription_details span.wrap label {
+	clear:both;
+}
+
 #wcsatt_data .subscription_scheme_data label {
 	margin: 0 0 0 -138px;
 	width: 118px;


franticpsyx:
Hey @rezzz-dev -- can you share a screenshot of the issue along with WC / WCS version numbers? Thanks!

assets/css/wcsatt-write-panels.css
```diff
@@ -115,6 +115,10 @@
 	float: left;
 }
 
+#wcsatt_data .subscription_scheme_data p._satt_subscription_details span.wrap label {
```
franticpsyx: This seems to be targeting `label.wcs_hidden_label` which should be hidden in WCS 2.1/2.2.

rezzz-dev:
screen shot 2017-06-15 at 10 44 11

That's what I'm seeing. The site I'm encountering this on is actually running WCS 2.0.22 I guess that's why.

franticpsyx:
@rezzz-dev 

Ah makes sense now! For SATT 2.0 the minimum required WC/WCS versions will be bumped to 2.6/2.1.

See https://github.com/Prospress/woocommerce-subscribe-all-the-things/blob/2.0/readme.txt

rezzz-dev:
I was hired to update their site, so stay tuned for any updates

PR #136, opened 2017-06-16
Product API Rewrite

### LINKED ISSUE #42: SATT Products API

### LINKED ISSUE #96: Support WC 3.0
Opening this to start working on WC 2.7 compatibility.

Since SATT does not extend any core CRUD classes, WC 2.7 compatibility is limited to the following tasks (more added as we investigate):

- Stop accessing CRUD object properties directly.
- Stop accessing deprecated object properties.
- Look for deprecated CRUD object functions and use new ones where applicable.
- Replace deprecated filters with new equivalents.
- Phase out use of post/post meta functions.
- Use CRUD API to save plugin data on product objects.

### LINKED ISSUE #102: SATT product in renewal after failed order is not converted to sub

### LINKED ISSUE #127: MnM - Subscription period is showing up as "array"
The subscription plan is $85 / array instead of $85 / month. 

image

This was a mix and match product with the following settings:

image

SATT 2.0 -alpha
WC-3.1 -alpha
MNM- 1.2.3
Subs - 2.2.7

Granted, I'm using the bleeding edge versions of pretty much everything, but just a heads up in case this persists.

franticpsyx:
In #42 we discussed the need to write a proper SATT Product API that would allow us to 

- get/set subscription scheme data, and
- get/set the active subscription scheme

directly on product objects. 

Conceptually, this is more in line with the SATT concept. Since the beginning, we _wanted_ to treat the subscription component of a product as something that:

- is decoupled from the product type,
- is applied on the product object,
- modifies the return value of various product methods, such as prices and price strings.

v1 was a proof of concept based around the de-coupling bit -- one that quickly grew much larger than originally intended. As such, v1 basically:

- _enhanced_ cart items with data that tricked WCS into treating them as subscription products; and 
- made some creative (ab)use of filters to modify the price strings displayed in the shop catalog and single-product pages.

v1:

- did not actually extend product objects with a subscription component; 
- did not provide a formal definition of a _subscription scheme_;
- did not provide an API for getting/setting subscription data on product objects;
- made it hard to understand what the actual _subscription data_ is (no API);
- did not provide a consistent (always on) mechanism for making product prices and other method return values dependent on the subscription scheme set _on_ product objects.

Our main goal with v2 was to work on these shortcomings. v2:

1. Addresses these issues by introducing a [Product API](https://github.com/Prospress/woocommerce-subscribe-all-the-things/blob/2.0/includes/class-wcsatt-product.php) that can be used to get/set subscription data _on_ product objects.
2. Provides a more formal (object) [representation](https://github.com/Prospress/woocommerce-subscribe-all-the-things/blob/2.0/includes/class-wcsatt-scheme.php) of the _subscription scheme_ concept, which is used to set/get scheme data and can be later extended to handle the retrieval/storage of the actual data from/to the DB, should we decide to store scheme content in a structure other than product meta.
3. Redefines the way product prices and other method return values are [mutated](https://github.com/Prospress/woocommerce-subscribe-all-the-things/blob/2.0/includes/class-wcsatt-scheme-prices.php) depending on the subscription scheme set _on_ product objects.
4. Introduces some basic unit tests for the Product API which also provide a good starting point for understanding the nature/scope of these changes.

It also: 

1. Streamlines the UI of the 'Product Data > Subscriptions' tab panel.
2. Adds WC 3.0 / WCS 2.2 support.
3. Brings many improvements to the SATT integration with Product Bundles and Composite Products.

Ideally, I would have liked each of these changes to be reviewed and merged separately -- however, the need to rewrite the entire codebase in order to utilize the Product API meant that some of these tasks would need to be done in multiple passes to be tacked in isolation from each other (e.g. finalize the refactor with WC 2.6 / WCS 2.1 in mind and then work on WC 3.0 / WCS 2.2 compatibility in a separate branch).

In the end, I took the (admittedly sloppy) route of doing everything in a 2.0 branch. Due to time restrictions, and as I knew I'd be doing the refactor myself (not an excuse), I decided to work on isolated parts of the codebase that I could hands-on test sequentially and in full (write and test the API, refactor the catalog/single-product parts of the codebase to utilize the API, move to the cart, rewrite the order-related code and finally revise the integrations).

The "right" approach would have been to split this work up in multiple issues and branches, and rely on unit tests to test each individual component/class. Unfortunately, unit tests can take a lot of time to write properly -- and for the last few months time has been a scarce resource at SW and Prospress. 

On a personal note, unless I have the resources to be really exhaustive/thorough and systematic when writing tests, I find that a better use of my time is often to write solid code and _immediately_ test each new/modified logical block against a real-world scenario. This is _also_ a time-consuming process, but I find it gives me the hands-on reassurance I need to move on when refactoring code.

### But this is huge! How do I review this?

Although it would have been a good idea to split this up into multiple branches/issues/PRs (API, shop catalog, cart, orders) it would **still** have been a pain to review the changes by looking at diffs. 

Remember we are not dealing with changes here -- but with a full rewrite. The entire logic behind SATT is different. **There is very little to compare between v1 / v2**. This doesn't make me any less guilty.

The best way to look at this PR is to treat `v2` as new code and ignore the diff.

* Start with the [Product API class](https://github.com/Prospress/woocommerce-subscribe-all-the-things/blob/2.0/includes/class-wcsatt-product.php) and read the inline docs/comments. It should be self-explanatory.
* To understand how product prices are coupled with the active subscription scheme set on product objects, check out the `WCS_Scheme_Prices` class [here](https://github.com/Prospress/woocommerce-subscribe-all-the-things/blob/2.0/includes/class-wcsatt-scheme-prices.php).
* Have a look at the [Scheme object](https://github.com/Prospress/woocommerce-subscribe-all-the-things/blob/2.0/includes/class-wcsatt-scheme.php).
* Move to the `WCS_ATT_Cart` [class](https://github.com/Prospress/woocommerce-subscribe-all-the-things/blob/2.0/includes/class-wcsatt-cart.php) to see how the Product API is being used in cart context.
* Finish with the `WCS_ATT_Order` [class](https://github.com/Prospress/woocommerce-subscribe-all-the-things/blob/2.0/includes/class-wcsatt-order.php) to see what kind of data SATT stores/retrieves on/from order item objects, and how it is used to restore that subscription state of a product object instantiated from an order item.

### Backwards compatibility

If you used any of the filters present in v1, you will find that these are still in use in v2, almost in their entirety. Additionally, scheme objects implement the `ArrayAccess` interface, so if you extended the v1 scheme data arrays with custom fields, your code should still work.

If you notice a change that looks breaking/unnecessary, and/or feel something is not backwards compatible, let's discuss it!

### Summing up

Some of the work could have been done in separate branches (e.g. the admin UI improvements), but I hope that the amount of inline documentation in the refactored code and the way it's organized should make it straightforward to understand what's changed in v2.

### What's next

1. This PR gets reviewed. Anyone who has written custom code or plugin sitting on top of SATT, feel free to take this for a spin @seb86 @javorszky .
2. If this is not a terrible mess, we clean things up based on the review and tag a 2.0.0-beta.
3. More testing and fixes.
4. We release SATT 2.0.0.

### Final notes

1. We need more unit tests to be written for: 

* Product API functions;
* Cart API functions and hooks;
* scheme objects;
* price-related mutations/filters coupled with the Product API (see `WCS_ATT_Scheme_Prices`).

2. We'll be documenting some of our intentions/plans in a mini Roadmap soon, which reflects the scarce (but increasing!) feedback we have received over the last couple years. It will be added to the [Wiki](https://github.com/Prospress/woocommerce-subscribe-all-the-things/wiki/Roadmap) later today. Noting this here as I find it's always a good idea to put things into perspective when discussing/reviewing major changes like this one.

closes #42 
closes #96 
closes #102 
closes #127


seb86:
I know i'm not a reviewer but it looks all good to me. I will be updating my add-ons soon. Great work @franticpsyx

franticpsyx:
Thanks @seb86 :)

Tried to add you in the list but couldn't. My options were limited to repo team members/collaborators for some reason (?). Will see if I can fix this, otherwise @thenbrent should be able to take care of this!

franticpsyx:
@seb86 added you!

franticpsyx:
@javorszky Added you as well!

javorszky:
Woop, thank you! I'll do a review of it before monday evening UK time (WooCommerce meetup). :)

includes/admin/meta-boxes/views/subscription-scheme.php
```diff
@@ -11,7 +11,10 @@
 }
```
thenbrent: Did you mean to change permissions on this file (from 755 to 644) or was that accidental?

franticpsyx:
@thenbrent let's tackle some of the smaller issues by discussing them here and see what we might need/want to discuss (or solve) in separate issues. Starting with your general comments:


> As mentioned in one comment, there are a number of large functions and classes that could be broken down to smaller ones [...]

In general I agree, but when looking at collections of tightly related, static functions, such as the `WCS_ATT_Product` API, I think what we'd ideally want is a way to re-structure the way the code is used/called, instead of the way the code is organized in files/directories. A pattern like this would be ideal, but I am not sure if it's possible:

`WCS_ATT_Product::Prices::get_price_html` and `WCS_ATT_Product::Schemes::get_subscription_scheme`

I wouldn't have any reservation if we were talking about splitting up an object into multiple smaller objects, but when discussing classes such as `WCS_ATT_Product`, is there any benefit in terms of testability or simplicity? My feeling in this case is that keeping all functions in one place makes it easier for devs to know where to look.

> Class names are all prefixed with WCS_ATT_, but file names use class-wcsatt- (no separation between WCS and ATT). That’s breaking WC file naming conventions. I know there is legacy here, but perhaps we could break from that with the 2.0 release?

👍 -- Will push a commit here and link to this comment.

> Consistent variable names, e.g. sometimes $subscription_schemes is used and other times just $schemes.

👍 -- always try to keep this in mind. Sometimes though there is a temptation to relax this requirement, e.g. when there is a need to declare "modified" versions of existing variables with long names. When you have both `my_long_variable_name` and `my_modified_long_variable_name` heavily used in a function, readability may become an issue. I often catch myself simplifying the name of a variable in a function block if I know that I will be declaring/using similarly named variables in complex assignments, as I find it's more important to be able to see all of my code without scrolling horizontally. I am not sure if there is a coding standard for this, but here my vote always went to readability vs naming consistency.

Not sure if I should re-scan the code and make changes though, as lack of unit tests coverage means the code is quite fragile to variable name changes at the moment.

> public methods can not be deleted safely while still maintaining backwardy compatibility. Instead they need to be deprecated by calling __deprecated_function() or wc_deprecated_function()). For example, if 3rd party code was calling WCS_ATT_Display::convert_to_sub_product_options(), which has now been replaced with WCS_ATT_Display::show_subscription_options(), it would trigger a fatal error when the store updates to 2.0. [...]

This is a big can of worms..! Ideally, hook callbacks should be clearly separated from functions/methods meant to be used as part of a public "API". I often find it tough to decide how to split up "event-related" code (such as filter callbacks) from "object methods" or static functions that are meant to be used publicly.

If it was possible, I would have liked to be able to declare specific hook callbacks as `private`, if they are not meant to be touched or used publicly. Not being able to limit the scope of callbacks is a shortcoming of WP and something we can only "work around" by adding `@access internal`, but I find that a bit moot for functions that are `public` anyway. The example you are bringing up is one of the few functions I didn't deprecate, and the reason is that it's one of those I don't want/expect anyone to unhook directly.

> moved a little bit of the PR description into the Wiki for posterity's sake once this PR is merged and long forgotten, it seemed like a good basis for a general Where do I start?! section for devs. See: https://github.com/Prospress/woocommerce-subscribe-all-the-things/wiki/Brief-Developer-Guide-to-Subscribe-all-the-Things

👍 Moving on to the review comments!

includes/class-wcsatt-product.php
```diff
@@ -0,0 +1,856 @@
...
+
+		// Allow WCS to recognize any product as a subscription.
+		add_filter( 'woocommerce_is_subscription', array( __CLASS__, 'filter_is_subscription' ), 10, 3 );
+
+		// Delete object meta in use by the application layer.
+		add_action( 'woocommerce_before_product_object_save', array( __CLASS__, 'delete_reserved_meta' ) );
+
+		// Make product prices scheme-dependent.
+		WCS_ATT_Scheme_Prices::add_price_filters();
+		WCS_ATT_Scheme_Prices::add_price_html_filters();
```
franticpsyx: @thenbrent I felt that adding these here gives developers a clearer impression of how _products_ are modified to support the subscription schemes concept. The Product API is meant for setting/getting subscription states directly -- at the same time, however, the subscription states available/set on product objects change the return values of some `WC_Product` methods via filters. Registering these filters here makes this clearer. I'm not sure if the best place for these callbacks is `WCS_ATT_Scheme_Prices` or a new `WCS_ATT_Product_Prices` class, but I wanted to keep them out of the API class.

includes/class-wcsatt-product.php
```diff
@@ -0,0 +1,856 @@
...
+		return ! empty( $active_scheme_key );
+	}
+
+	/**
+	 * Determines if the product can be purchased on a recurring basis.
+	 *
+	 * @param  WC_Product  $product  Product object to check.
+	 * @return boolean               Result of check.
+	 */
+	public static function has_subscriptions( $product ) {
```
franticpsyx: 👍 Opened #150

includes/class-wcsatt-product.php
```diff
@@ -0,0 +1,856 @@
...
+	 * Determines if the product is purchasable on a recurring basis only.
+	 *
+	 * @param  WC_Product  $product  Product object to check.
+	 * @return boolean               Result of check.
+	 */
+	public static function has_forced_subscription( $product ) {
+
+		if ( '' === ( $forced = self::get_product_property( $product, 'has_forced_subscription' ) ) && self::has_subscriptions( $product ) ) {
+
+			$forced = WCS_ATT_Core_Compatibility::is_wc_version_gte_2_7() ? $product->get_meta( '_wcsatt_force_subscription', true ) : get_post_meta( WCS_ATT_Core_Compatibility::get_id( $product ), '_wcsatt_force_subscription', true );
```
franticpsyx: In most cases, you will notice that the raw data obtained via `get_meta` or `get_post_meta` requires some kind of transformation -- then the approach followed here is to store/cache it on the object using `set_product_property`.

Two examples:

1. "Raw" subscription scheme data is used to create scheme objects once. The created objects are stored/cached on the product as "runtime meta".

2. "Raw" subscription scheme data only includes the schemes defined at product level, but cart-level schemes can be appended to product objects in the cart after loading cart session data. 

So `get/set_product_property` is basically used as a "runtime meta" getter/setter, which does some back-compat handling.

Also see https://github.com/Prospress/woocommerce-subscribe-all-the-things/pull/136#discussion_r123648135

includes/class-wcsatt-product.php
```diff
@@ -0,0 +1,856 @@
...
+
+			// Attempt to get opion from parent if undefined on variation.
```
franticpsyx: 👍 Added to https://github.com/Prospress/woocommerce-subscribe-all-the-things/issues/151

includes/class-wcsatt-product.php
```diff
@@ -0,0 +1,856 @@
...
+		return 'yes' === $forced;
+	}
+
+	/**
+	 * Checks a product object to determine if it is a WCS subscription-type product.
+	 *
+	 * @param  WC_Product  $product  Product object to check.
+	 * @return boolean               Result of check.
+	 */
+	public static function is_legacy_subscription( $product ) {
```
franticpsyx: 👍 Added to #151

includes/class-wcsatt-product.php
```diff
@@ -0,0 +1,856 @@
...
+
+		$active_scheme_key = self::get_subscription_scheme( $product );
+		$scheme_key        = '' === $scheme_key ? $active_scheme_key : $scheme_key;
+
+		// Attempt to switch scheme when requesting the price html of a scheme other than the active one.
+		$scheme_switch_required = $scheme_key !== $active_scheme_key;
+		$switched_scheme        = $scheme_switch_required ? self::set_subscription_scheme( $product, $scheme_key ) : false;
+
+		// Scheme switch required but unsuccessful? Problem. Set price html to an empty string.
+		if ( $scheme_switch_required && false === $switched_scheme ) {
```
franticpsyx: I think `self::set_subscription_scheme` returns `false` or `true` -- `false` would be returned if you request a switch to a scheme with a non-existent key.

includes/class-wcsatt-product.php
```diff
@@ -0,0 +1,856 @@
...
+				}
+
+				$html_from_text = WCS_ATT_Core_Compatibility::get_price_html_from_text( $product );
+
+				if ( self::has_forced_subscription( $product ) ) {
+
+					// No infinite loops, thank you.
+					WCS_ATT_Scheme_Prices::remove_price_html_filters();
+					$price_html = $product->get_price_html();
+					WCS_ATT_Scheme_Prices::add_price_html_filters();
```
franticpsyx: 👍 Added to #151

includes/class-wcsatt-product.php
```diff
@@ -0,0 +1,856 @@
...
+								$use_discount_html_format = true;
+								if ( $discount !== $scheme->get_discount() ) {
+									if ( '' === $discount ) {
+										$discount = $scheme->get_discount();
+									} else {
+										$has_variable_discount = true;
+									}
+								}
+							}
+						}
```
franticpsyx: This basically loops through all schemes to decide two things:

1. Whether to use a "discount-based" html format.
2. Whether different schemes have varying discounts.

We could create 2 dedicated functions for these, but they wouldn't be useful in any other place.

Large, deeply nested functions are evil indeed. Maybe it's just me and my scripting language background, but string-generating functions like this one are a typical example of something that could become more complex to review/understand if you split it up too much or use objects to do the same thing.

includes/class-wcsatt-product.php
```diff
@@ -0,0 +1,856 @@
...
+
+	/**
+	 * Returns the sale price per subscription period.
+	 *
+	 * @param  WC_Product  $product     Product object.
+	 * @param  string      $scheme_key  Optional key to get the price of a specific scheme.
+	 * @param  string      $context     Function call context.
+	 * @return mixed                    The sale price charged per subscription period.
+	 */
+	public static function get_sale_price( $product, $scheme_key = '', $context = 'view' ) {
```
franticpsyx: 👍 Opened https://github.com/Prospress/woocommerce-subscribe-all-the-things/issues/152

includes/class-wcsatt-product.php
```diff
@@ -0,0 +1,856 @@
...
+	*/
+
+	/**
+	 * Property getter (compatibility wrapper).
+	 *
+	 * @param  WC_Product  $product   Product object.
+	 * @param  string      $property  Property name.
+	 * @return mixed
+	 */
+	private static function get_product_property( $product, $property ) {
```
franticpsyx: Agreed -- see https://github.com/Prospress/woocommerce-subscribe-all-the-things/pull/136#discussion_r125583798

As it stands, `get_runtime_meta` would be the right name for it, if we end up using it after all. Looks like we might need/want to take the discussion about these in a separate issue?

includes/class-wcsatt-product.php
```diff
@@ -0,0 +1,856 @@
...
+	/**
+	 * Delete object meta in use by the application layer.
+	 *
+	 * @param  WC_Product  $product
+	 */
+	public static function delete_reserved_meta( $product ) {
+		$reserved_meta_keys = array( 'has_forced_subscription', 'subscription_schemes', 'active_subscription_scheme_key', 'default_subscription_scheme_key' );
+		foreach ( $reserved_meta_keys as $reserved_meta_key ) {
+			$product->delete_meta_data( '_' . $reserved_meta_key );
+		}
```
franticpsyx: Never done this before. It's the first time I actually need to store "runtime" data on product objects. Using the meta data array to temporarily store "stuff" related to the object is something I have seen in core, but I am having troubling finding where it was. Probably order items? Of course if core devs do something, that doesn't necessarily make it a good example to follow 😄 

Realistically, we have no other way to store "runtime" data on product objects. What we have done here looks safe -- might have been better to use a single meta key to store all data. With that said, I agree that using CRUD to do it is probably a bad idea.

> I don't have a great alternative to suggest, but off the top of my head, it might make sense to store the values in `WCS_ATT_Product` instead.

This would definitely _work_. We could map data to objects using object instance IDs. It still feels more natural to me that object-specific data should be stored _on_ the object. I am really torn here, @thenbrent .

franticpsyx: PS: Actually I just realized I have done something similar in PB/CP, not that it matters.

assets/js/wcsatt-single-add-to-cart.js
```diff
@@ -31,7 +31,7 @@
 				var $scheme_option = $( this ),
 					scheme_data    = $( this ).find( 'input' ).data( 'custom_data' );
 
-				bundle.satt_schemes.push( { el: $scheme_option, data: scheme_data } );
+				bundle.satt_schemes.push( { el: $scheme_option, data: scheme_data, price_html_o: $scheme_option.find( '.subscription-price' ).html(), details_html: $( '<div>' ).html( $scheme_option.find( '.subscription-details' ) ).html() } );
```
franticpsyx: `_o` stands for "original". The original/initial price html in the subscription options list is replaced by a "live" version, which is calculated based on the chosen options/configuration of a bundle/composite.

e.g. "From $X" could become "$Y" while/after configuring a bundle.

includes/class-wcsatt-core-compatibility.php
```diff
@@ -48,7 +55,7 @@ private static function get_wc_version() {
 	 * @return boolean
 	 */
 	public static function is_wc_version_gte_2_7() {
```
franticpsyx: These `_MAJOR_MINOR` functions are probably unnecessary and could be deprecated/removed. Are we dropping support for WC 2.6, after all? If so, then we don't need them anyway.

includes/class-wcsatt-scheme-prices.php
```diff
@@ -1,388 +1,484 @@
...
+	 * @return  void
 	 */
-	public static function filter_get_variation_prices( $prices_array, $product, $display ) {
+	public static function add_price_html_filters() {
 
-		$subscription_scheme = self::$price_overriding_scheme;
+		add_filter( 'woocommerce_get_price_html', array( __CLASS__, 'filter_get_price_html' ), 9999, 2 );
 
-		if ( $subscription_scheme && self::has_subscription_price_override( $subscription_scheme ) ) {
+		do_action( 'wcsatt_add_price_html_filters' );
```
franticpsyx: When the time comes to merge (the core parts of) SATT into WCS, we will need to:

1. Run some migration code on the WCS side. Some new data/meta will be created, e.g. as "legacy" subscription product types are converted to "core" types. The new data will be (most likely) prefixed with `_wcs_` -- e.g. `_wcs_subscription_schemes` vs. `_wcsatt_schemes` in SATT v2. We can be as liberal as we want with choosing new names for **meta and filters**.

2. Run some migration code on the SATT side to prepare SATT meta from merging into WCS, e.g. convert `_wcsatt_schemes` meta to `_wcs_subscription_schemes`. Add a **filter deprecator** class in SATT as the only "remnant" of all code ported into WCS, to avoid the overhead of having all of it in WCS. People who used to run SATT before and need full backwards compatibility with WCS 3.0 will need to run the latest SATT version, which will (a) migrate their data and (b) deprecate old filters.

Part (2) effectively **makes SATT fully responsible for the migration of existing SATT data for compatibility with WCS 3.0**, not the other way around. The interaction of WCS 3.0 with SATT should be limited to checking whether it exists and show an admin notice when the installed version is old.

Summing up:

Because lots of DB data already exists in SATT that will need to be migrated,

and;

assuming that the best approach is to eventually make SATT responsible for migrating its own data,

I'd vote in favor of keeping all filter/meta names in SATT agnostic to the eventual "merge" with WCS, to avoid making _any_ migration-related decisions now, and maintain naming consistency.

includes/class-wcsatt-scheme-prices.php
```diff
@@ -1,388 +1,484 @@
...
-				asort( $sale_prices );
+				$prices[ 'regular_price' ] = $subscription_scheme->get_regular_price();
+				$prices[ 'sale_price' ]    = $subscription_scheme->get_sale_price();
 
-				$prices_array = array(
-					'price'         => $prices,
-					'regular_price' => $regular_prices,
-					'sale_price'    => $sale_prices
-				);
+				$prices[ 'price' ] = '' !== $prices[ 'sale_price' ] && $prices[ 'sale_price' ] < $prices[ 'regular_price' ] ? $prices[ 'sale_price' ] : $prices[ 'regular_price' ];
```
franticpsyx: 👍 Added in #151

includes/class-wcsatt-scheme-prices.php
```diff
@@ -1,388 +1,484 @@
...
-		return $price;
+		return $price_filter_exists;
 	}
 
 	/**
-	 * Filter get_regular_price() calls to take price overrides into account.
+	 * Returns the "base" subscription scheme by finding the one with the lowest recurring price.
+	 * If prices are equal, no interval-based comparison is carried out:
+	 * Reason: In some applications "$5 every week for 2 weeks" (=$10) might be seen as "cheaper" than "$5 every month for 3 months" (=$15), and in some the opposite.
+	 * Instead of making guesswork and complex calculations, we can let scheme order be used to define the "base" scheme manually.
```
franticpsyx: Since the "scheme options" admin UI allows users to sort them, I thought it might be best to let sort order imply which scheme is the "base" one when all recurring prices are equal.

Note sure how/if we should revise this when we introduce trials and sign up fees. Personally I'd keep it simple and still rely on sort order to define the "base".

includes/class-wcsatt-scheme.php
```diff
@@ -0,0 +1,431 @@
...
+			}
+
+			if ( 'inherit' === $this->data[ 'pricing_mode' ] ) {
+				$this->data[ 'discount' ] = isset( $args[ 'data' ][ 'subscription_discount' ] ) ? wc_format_decimal( $args[ 'data' ][ 'subscription_discount' ] ) : '';
+			}
+		}
+
+		$this->data[ 'context' ] = isset( $args[ 'context' ] ) ? strval( $args[ 'context' ] ) : 'product';
+
+		$this->key = implode( '_', array_filter( array( $this->data[ 'interval' ], $this->data[ 'period' ], $this->data[ 'length' ] ) ) );
```
franticpsyx: Yep, we'll need to change this when we introduce trials/sign-up fees.

This is one reason why keys are generated on object creation and not stored in the DB. As we add more features, keys will need to change.

There might be some back-compat challenges, but let's think about them when we add those features in (e.g. when no trial / sign-up fee exists, omit all trial/sign-up data from the key).

includes/class-wcsatt-scheme.php
```diff
@@ -0,0 +1,431 @@
...
+	}
+
+	/**
+	 *
+	 * Returns a unique scheme identifier. For now it just returns the key. Later it could be a CRUD object identifier.
+	 *
+	 * @return string  A unique identifier.
+	 */
+	public function get_id() {
```
franticpsyx: I don't think `get_id()` is used anywhere. Agreed it's better to remove it. Added to #151

> Then if it moves to extend `WC_Data`, we could deprecate `get_key()` and provide `get_id()` without changing its return value.

In this scenario, CRUD IDs would be most likely BIGINT, while keys would still be used as string representations of schemes. Doesn't matter at this time but I have the impression we'd need to keep `get_key`.

includes/class-wcsatt-scheme.php
```diff
@@ -0,0 +1,431 @@
...
+	|--------------------------------------------------------------------------
+	*/
+
+	/**
+	 * Returns the date on which the subscription scheme will expire,
+	 * based on the subscription's length and calculated from either the $from_date if specified, or the current date/time.
+	 *
+	 * @param  mixed  $from_date  A MySQL formatted date/time string from which to calculate the expiration date, or empty (default), which will use today's date/time.
+	 */
+	public function get_expiration_date( $from_date = '' ) {
```
franticpsyx: 👍 Added to #151

includes/class-wcsatt-scheme.php
```diff
@@ -0,0 +1,431 @@
...
+		return $expiration_date;
+	}
+
+	/**
+	 * Returns the date on which the subscription scheme trial will expire,
+	 * based on the subscription's length and calculated from either the $from_date if specified, or the current date/time.
+	 *
+	 * @param  mixed  $from_date  A MySQL formatted date/time string from which to calculate the trial expiration date, or empty (default), which will use today's date/time.
+	 */
+	public function get_trial_expiration_date( $from_date = '' ) {
```
franticpsyx: Same as above.

includes/class-wcsatt-schemes.php
```diff
@@ -1,317 +1,166 @@
...
+	*/
 
 	/**
-	 * Returns the active subscription scheme of a cart item, or false if the cart item is a one-off purchase.
+	 * Returns a subscription scheme by id. Legacy.
 	 *
 	 * @return array
 	 */
-	public static function get_active_subscription_scheme( $cart_item ) {
+	public static function get_subscription_scheme_by_id( $id, $schemes ) {
```
franticpsyx: I think this function is a leftover from v1 and not used in v2 at all. In v1, "keys" were called "ids". This was changed in v2 to reflect the notion that "id" usually refers to a DB identifier.

Note that some filter names still use the `_id` suffix for compat with v1. I didn't touch those to avoid deprecating them.

templates/single-product/satt-product-options.php
```diff
@@ -28,10 +28,10 @@
 	}
 
 	?><ul class="wcsatt-options-product"><?php
-		foreach ( $options as $option_id => $option ) {
-			?><li class="<?php echo $option_id !== '0' ? 'subscription-option' : 'one-time-option'; ?>">
+		foreach ( $options as $option ) {
+			?><li class="<?php echo $option[ 'value' ] !== '0' ? 'subscription-option' : 'one-time-option'; ?>">
 				<label>
-					<input type="radio" name="convert_to_sub_<?php echo $product->id; ?>" data-custom_data="<?php echo esc_attr( json_encode( $option[ 'data' ] ) ); ?>" value="<?php echo $option_id; ?>" <?php checked( $option[ 'selected' ], true, true ); ?> />
+					<input type="radio" name="convert_to_sub_<?php echo $product_id; ?>" data-custom_data="<?php echo esc_attr( json_encode( $option[ 'data' ] ) ); ?>" value="<?php echo esc_attr( $option[ 'value' ] ); ?>" <?php checked( $option[ 'selected' ], true, true ); ?> />
```
franticpsyx: Added to #151

includes/class-wcsatt-order.php
```diff
@@ -0,0 +1,153 @@
...
+ *
+ * @class    WCS_ATT_Order
+ * @version  2.0.0
+ */
+class WCS_ATT_Order {
+
+	public static function init() {
+
+		// Restore subscription data when creating a cart item using an order item as reference.
+		add_filter( 'woocommerce_order_again_cart_item_data', array( __CLASS__, 'restore_cart_item_from_order_item' ), 10, 3 );
```
franticpsyx: Should do the trick in combination with https://github.com/Prospress/woocommerce-subscribe-all-the-things/commit/bf7902962f722308eaee69ad548445423210870f

includes/class-wcsatt-order.php
```diff
@@ -0,0 +1,153 @@
...
+	|--------------------------------------------------------------------------
+	*/
+
+	/**
+	 * Returns the key of the subscription scheme applied on the product when it was purchased.
+	 *
+	 * @param  array  $order_item
+	 * @return string|false|null
+	 */
+	public static function get_subscription_scheme( $order_item ) {
```
franticpsyx: This is for consistency with `WCS_ATT_Cart::get_subscription_scheme` (key only) and `WCS_ATT_Product::get_subscription_scheme` (key or object).

Maybe we could make all of them return the key or object? Or we could rename both this and `WCS_ATT_Cart::get_subscription_scheme`? 

Note there are more functions that operate with keys which don't include "key" in their name, e.g. `WCS_ATT_Cart::set_cart_subscription_scheme`.

includes/class-wcsatt-order.php
```diff
@@ -0,0 +1,153 @@
...
+	/**
+	 * Attempts to restore the subscription state of a product instantiated using an order item as reference.
+	 *
+	 * @param  WC_Product  $product
+	 * @param  array       $order_item
+	 * @return WC_Product
+	 */
+	public static function restore_product_from_order_item( $product, $order_item ) {
+		if ( null !== ( $scheme_key = self::get_subscription_scheme( $order_item ) ) ) {
+			WCS_ATT_Product::set_subscription_scheme( $product, $scheme_key );
```
franticpsyx: > If that's the behaviour, we'll probably want to prevent renewal/resubscribe to a product when the scheme has been deleted to avoid any confusion with attempting to resubscribe and the product no longer creating a subscription. Similar to the way you can't resubscribe to a subscription product that has since been deleted (but you can remain subscribed to it).

👍 I think that's the best approach. Opened https://github.com/Prospress/woocommerce-subscribe-all-the-things/issues/153

includes/class-wcsatt-order.php
```diff
@@ -0,0 +1,153 @@
...
+	 * @param  array          $cart_item
+	 * @return void
+	 */
+	public static function save_subscription_scheme_meta( $order_item, $cart_item_key, $cart_item ) {
+
+		$scheme_key = WCS_ATT_Cart::get_subscription_scheme( $cart_item );
+
+		if ( null !== $scheme_key ) {
+			$scheme_key = false === $scheme_key ? '0' : $scheme_key;
+			$order_item->add_meta_data( '_wcsatt_scheme', $scheme_key, true );
```
franticpsyx: The approach I'd follow would be to start fresh with data/hook names in WCS 3.0, and expect existing SATT users to install a WCS 3.0-compatible version that will:

- migrate their data; and
- include a hook deprecator.

For details, see https://github.com/Prospress/woocommerce-subscribe-all-the-things/pull/136#discussion_r125601822

includes/class-wcsatt-integrations.php
```diff
@@ -348,212 +379,252 @@ public static function is_bundle_type_product( $product ) {
...
+				self::set_bundled_product_subscription_schemes( $product, $composite );
+			}
+
+			if ( $product = $component_option->get_product( array( 'having' => 'regular_price', 'what' => 'min' ) ) ) {
+				self::set_bundled_product_subscription_schemes( $product, $composite );
+			}
+
+			if ( $product = $component_option->get_product( array( 'having' => 'regular_price', 'what' => 'max' ) ) ) {
+				self::set_bundled_product_subscription_schemes( $product, $composite );
+			}
```
franticpsyx: 👍 Added to #151

includes/admin/meta-boxes/views/subscription-scheme.php
franticpsyx: See https://github.com/Prospress/woocommerce-subscribe-all-the-things/pull/148

I think that `filemode` was accidentally set to `true` in one of the SATT repos I've been doing most of the work so far.

With `filemode` set to `false`, files should be 644 and directories 755 -- right?

thenbrent:
> A pattern like this would be ideal, but I am not sure if it's possible:
> WCS_ATT_Product::Prices::get_price_html and 
> WCS_ATT_Product::Schemes::get_subscription_scheme

Something like that is possible using static properties with other classes, e.g.

```php
class WCS_ATT_Product {

	public static $prices;
	public static $schemes;

	public function init() {
		self::$prices  = new WCS_ATT_Product_Prices();
		self::$schemes = new WCS_ATT_Product_Schemes();
	}

	public function some_function() {
		$price_html = self::$prices::get_price_html();
		$schemes = self::$schemes::get_subscription_scheme();
	}
}

```

That still requires the restructuring of files/folders though, which it sounds like you want to avoid.

> I wouldn't have any reservation if we were talking about splitting up an object into multiple smaller objects, but when discussing classes such as WCS_ATT_Product, is there any benefit in terms of testability or simplicity? My feeling in this case is that keeping all functions in one place makes it easier for devs to know where to look.

Knowing the right class doesn't necessarily mean being able to find them easily though (speaking from experience with some of the unfortunately large Subscriptions classes I created, like `WC_Subscriptions_Switcher`, which would be much easier to trace code if it was split into 4 or 5 smaller classes).

It's a tough tradeoff though (and not a blocking issue here).

FWIW, WC does recommend not creating "God Objects" too: https://docs.woocommerce.com/document/create-a-plugin/#section-22

---

> I am not sure if there is a coding standard for this, but here my vote always went to readability vs naming consistency.

The most relevant standard is this part:

> Don’t abbreviate variable names unnecessarily; let the code be unambiguous and self-documenting.

There's nothing about consistency across usage though. It's just been very handy when possible (e.g. when 3.0 comes around and you need to replace _all_ property access on all objects, it's much easier to search for `order->` or `product->`).

> Not sure if I should re-scan the code and make changes though, as lack of unit tests coverage means the code is quite fragile to variable name changes at the moment.

2.0 is a good chance I figure to make that change. But it's not a blocking issue. :)

> This is a big can of worms..! Ideally, hook callbacks should be clearly separated from functions/methods meant to be used as part of a public "API".

> If it was possible, I would have liked to be able to declare specific hook callbacks as private, if they are not meant to be touched or used publicly. Not being able to limit the scope of callbacks is a shortcoming of WP (PHP) and something we can only "work around" by adding @access internal, but I find that a bit moot for functions that are public anyway. The example you are bringing up is one of the few functions I didn't deprecate, and the reason is that it's one of those I don't want/expect anyone to unhook directly.

For sure, but unfortunately we're stuck with that limitation and have to work with it. Once the (callback) methods are public, it's possible they could be being used by 3rd parties, so I normally play it safe and deprecate them.

I don't mind if you want to play it more dangerously, especially with the smaller user base of SATT atm. It's definitely nice not having bottom of classes littered with old, unused code anymore.

includes/class-wcsatt-product.php
thenbrent: That makes sense, though I wonder if that means the `add_price_filters` and `add_price_html_filters` should actually be in this class too, and calling the corresponding `WCS_ATT_Scheme_Prices` method. That would provide a clean separation of concerns.

includes/class-wcsatt-product.php
thenbrent: > So get/set_product_property is basically used as a "runtime meta" getter/setter 

Ah that makes sense.

includes/class-wcsatt-product.php
thenbrent: > As it stands, get_runtime_meta would be the right name for it, if we end up using it after all.

I think that's a good name. Definitely a new convention (I've not seen a name like that before in any WC code), but it is suitable and accurately describes the purpose of the function.

includes/class-wcsatt-product.php
thenbrent: Ignore me, I think I was looking at `set_subscription_schemes()` not `set_subscription_scheme()` (i.e. the plural). 🤦‍♂️

includes/class-wcsatt-order.php
seb86: Would the product update all active subscriptions with the scheme that was deleted to expire or when it tries to renew the customers subscription?

assets/js/wcsatt-single-add-to-cart.js
thenbrent: > `_o` stands for "original". 

Could the key be changed to `price_html_original` or `price_html_unmodified` then? :)

includes/class-wcsatt-product.php
thenbrent: > This would definitely work. We could map data to objects using object instance IDs. It still feels more natural to me that object-specific data should be stored on the object.

Yeah, I can understand that.

One advantage (or disadvantage depending on usage) of storing it is `WCS_ATT_Product` is that the same data will be available for other instances of the product. For example, if some other code only has the product ID, and instantiates a new instance of that product, should it also get the same `has_forced_subscription`, `subscription_schemes`, `active_subscription_scheme_key` and `default_subscription_scheme_key` values as the other instance or instances of the product? 

Answering that will answer where the data belongs IMO.

includes/class-wcsatt-core-compatibility.php
thenbrent: > Are we dropping support for WC 2.6, after all?

I'm keen to, but letting you make the call on what you're most comfortable doing. :)

includes/class-wcsatt-scheme-prices.php
thenbrent: That's a good way to approach it. I was thinking we'd avoid having SATT installed if possible, but it makes more sense to have installed and exclusively responsible for backward compatibility rather than running that code in Subscriptions.

> I'd vote in favor of keeping all filter/meta names in SATT agnostic to the eventual "merge" with WCS, to avoid making any migration-related decisions now

Sounds good to me 👍

includes/class-wcsatt-order.php
thenbrent: @seb86 no, all subscriptions remain unchanged, the same way if you change a subscription product's billing schedule, it doesn't update all active subscriptions with that product as a line item. The subscription is a previously agreed to set of terms, changing the products, which is an offer to agree to a new set of terms, shouldn't change existing subscriptions by default. :)

includes/class-wcsatt-order.php
thenbrent: IMO, getters which explicitly mention the expected return value reduce ambiguity so are preferred. But I'll leave the decision whether to change both `WCS_ATT_Product::get_subscription_scheme` and `WCS_ATT_Cart::get_subscription_scheme` up to you. :)

franticpsyx:
@thenbrent I have taken some time to review this PR myself. There are few issues with it that need to be addressed:

1. First, regarding `WCS_ATT_Product`:

> That still requires the restructuring of files/folders though, which it sounds like you want to avoid.

It **is** a big class. After some consideration, it feel it would be better to break it up in 3 components:

* `WCS_ATT_Product`;
* `WCS_ATT_Product_Schemes`; and
* `WCS_ATT_Product_Prices`.

`WCS_ATT_Product` will include the schemes/prices API components. Following this re-structuring, I feel that the bulk of filters currently in `WCS_ATT_Scheme_Prices` are bit disconnected from `WCS_ATT_Product_Prices`, which seems like a more appropriate place. Even better, all price filters could be added in a `WCS_ATT_Product_Price_Filters` class, which in turn could be included by `WCS_ATT_Product_Prices`. Then, `WCS_ATT_Product_Price_Filters::add()` and `WCS_ATT_Product_Price_Filters::remove()` would replace the equivalent functions currently in `WCS_ATT_Scheme_Prices`.

A new `product` dir could be added to "house" these new classes and make the codebase easier to navigate. It's quite a bit of shuffling around, but it's better to get it done now.

2. Following this, `WCS_ATT_Schemes` and `WCS_ATT_Scheme_Prices` will basically contain deprecated functions and can be moved away in a `legacy` dir.

A second look at `WCS_ATT_Schemes` gives me the impression that the few functions left there (some of them introduced/renamed in v2) belong in the new classes we added in v2:

* `WCS_ATT_Schemes::get_scheme_prices( $raw_prices, $subscription_scheme )` should become `WCS_ATT_Scheme::get_prices( $raw_prices )` and called on the scheme object: `$subscription_scheme->get_prices( $raw_prices )`.
* `WCS_ATT_Schemes::price_filter_exists` actually belongs in `WCS_ATT_Product_Schemes`.
* `WCS_ATT_Schemes::get_base_scheme` would be happier as `WCS_ATT_Product_Schemes :: get_base_subscription_scheme`.
* Finally, `WCS_ATT_Schemes ::get_discounted_scheme_price` belongs in `WCS_ATT_Scheme`.

Basically the entire `WCS_ATT_Schemes` and `WCS_ATT_Scheme_Prices` classes are v1 remnants -- the Product API and Scheme object type introduced in v2 are better homes for the few functions currently left in there.

Will open a separate PR with these changes in a bit.

franticpsyx:
Oh and:

3. The `WCS_ATT_Display_Cart` class (cart template hooks) was born out of a quite big/messy v1 `WCS_ATT_Display` class. It might be a good idea to finish this off by creating a `WCS_ATT_Display_Product` class (single-product template hooks).

franticpsyx:
@thenbrent the last few commits in `2.0` should address the last bits of our discussion here.

Added #161 for reference and took the liberty to merge #162 myself. When you find a minute or two, could you please give me a sanity check for the following:

* Have a look at the directory structure [here](https://github.com/Prospress/woocommerce-subscribe-all-the-things/tree/2.0/includes) (`WCS_ATT_Product` components in `/product/` and `WCS_ATT_Display` components in `/display/`).
* See what's left in `WCS_ATT_Product` [here](https://github.com/Prospress/woocommerce-subscribe-all-the-things/blob/2.0/includes/class-wcs-att-product.php).
* Let me know if you find [this pattern](https://github.com/Prospress/woocommerce-subscribe-all-the-things/blob/2.0/includes/class-wcs-att-product.php#L23:L56) meaningful. It should keep things a bit cleaner/safer as action/filter hook callbacks are added in their own function and can only be registered once.

With (hopefully) all shuffling about complete, it should be easier to focus on the remaining `2.0` issues we opened after your review.

seb86:
Changes look good so far from just reading the code. Looks cleaner. @franticpsyx @thenbrent I will do a deeper review when I work on updating my feature extensions for SATT over the weekend.

PR #148, opened 2017-07-05
[2.0] Rename filenames to match classnames

franticpsyx:
See https://github.com/Prospress/woocommerce-subscribe-all-the-things/pull/136#pullrequestreview-45871671

@thenbrent Changed file names instead of class names for back-compat.

I can see some odd 755 permissions, not sure where they came from. I have set git to ignore permissions for this repo, so files should have been 644. Maybe the repo in my office dev machine is not configured correctly? Will need to look into this when I get back.


franticpsyx:
Also I had a look at template names -- they are all `satt-xxx-options.php`.

Not sure if it's a good idea, but we could change those to `xxx-subscription-options.php`: `product-subscription-options`, `cart-item-subscription-options.php` and `cart-subscription-options.php`.

thenbrent:
> we could change those to

I think that's probably a good idea. Better to break compatibility to them now with a smaller userbase than in a year or two when they get moved into Subscriptions core.

I'll merge this one now though and those can be done in a separate PR for posterity. 👍

PR #162, opened 2017-07-09
[2.0] Break up God API class and clean up old Scheme classes

### LINKED ISSUE #161: [2.0] Break up God API class and clean up old Scheme classes
See https://github.com/Prospress/woocommerce-subscribe-all-the-things/pull/136#issuecomment-313195277 and https://github.com/Prospress/woocommerce-subscribe-all-the-things/pull/136#issuecomment-313871639

franticpsyx:
Closes #161


includes/class-wcs-att-cart.php
```diff
@@ -20,7 +20,29 @@
  */
 class WCS_ATT_Cart {
 
+	/**
+	 * Flag to ensure hooks can be added only once.
+	 * @var bool
+	 */
+	private static $added_hooks = false;
```
thenbrent: This system should be unnecessary. WordPress will only ever add the same callback with the same priority once, see `_wp_filter_build_unique_id()` and `WP_Hook::add_filter()`.

includes/class-wcs-att-display.php
```diff
@@ -112,107 +103,9 @@ public static function show_subscription_options() {
...
-
-		return $content;
-	}
-
-	/**
-	 * Override the WooCommerce "Add to Cart" button text with "Sign Up Now".
-	 *
-	 * @since 1.1.1
-	 */
-	public static function add_to_cart_text( $button_text ) {
```
thenbrent: Did you intend to remove this method also? The deprecation notice below is only for the `get_subscription_options_content()` method.

includes/class-wcs-att-cart.php
franticpsyx: Always something to learn. Was probably over-zealous here. Will revert that check. 👍

includes/class-wcs-att-display.php
franticpsyx: Took the less safe approach of moving filter callbacks without deprecating them: `get_subscription_options_content` is not attached to a filter and looks more like something third-party code might be using. 

It's hard to draw a line but as we discussed I have "only" deprecated all (hopefully) public methods, excluding methods attached to filters/actions.

After putting all variables on a scale this seems like the "sane" thing to do: Price filter callbacks in v1 were attached/detached conditionally (not always-on), and some display callbacks return slightly different data in v2 (enough to cause issues in v2 templates, which have also been bumped).

PR #165, opened 2017-07-10
[2.0] Runtime meta handling

franticpsyx:
@thenbrent you might want to have a quick look at this one as it's right at the core of SATT.

Checklist:

1. None of the [meta](https://github.com/Prospress/woocommerce-subscribe-all-the-things/compare/2.0...issue-163#diff-1f9b8d9552b9a679b925a590e66a072bR33) currently written to the DB by WCS should be persisted when saving a non-subscription-type product. See [this](https://github.com/Prospress/woocommerce-subscribe-all-the-things/compare/2.0...issue-163#diff-1f9b8d9552b9a679b925a590e66a072bR142).
2. For posterity, none of those meta fields are needed by WCS 3.0 (?) -- subscription details can be retrieved from scheme objects, which are created at runtime and saved in "runtime meta" territory. See `WCS_ATT_Product_Schemes::get_subscription_schemes`.
3. All "runtime data" has been consolidated under a single WC meta field, `_satt_data`, which is deleted when a product object is saved. See `WCS_ATT_Product::get/set_runtime_meta`.
4. One implication of this is that **runtime data is lost when a product is saved**, along with the subscription state of a product object. I am not sure if this is a bad idea -- I cannot imagine any scenario where `WC_Product::save` would be called that requires subscription state to be maintained after saving. (Need to think/check further.)

(4) is probably the most "critical" item in this list. If we don't want this behavior, then we will need to revise the way we associate subscription scheme/state data with product instances, and probably abandon the "runtime meta" approach.

### DIFF
--- a/includes/class-wcs-att-product.php
+++ b/includes/class-wcs-att-product.php
@@ -27,6 +27,19 @@ class WCS_ATT_Product {
 	private static $added_hooks = false;
 
 	/**
+	 * DB meta expected by WCS that needs to be added by SATT at runtime.
+	 * @var array
+	 */
+	private static $subscription_product_type_meta_keys = array(
+		'subscription_period',
+		'subscription_period_interval',
+		'subscription_length',
+		'subscription_trial_period',
+		'subscription_trial_length',
+		'subscription_sign_up_fee'
+	);
+
+	/**
 	 * Include Product API price and scheme components and add hooks.
 	 */
 	public static function init() {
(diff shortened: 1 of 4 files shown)


franticpsyx:
PS: I will need approval to merge this one.

thenbrent:
LGTM. I'm really not sure if this or the alternative approach of caching this data in memory within something else, like `WCS_ATT_Product` or similar is the best approach. But I think this is a good start and shipping it now allows us to see how it goes in SATT before having to make the call on how we do it in Subscriptions core. 👍

PR #173, opened 2017-07-13
Handling of renewals/resubscriptions

### LINKED ISSUE #153: [2.0] Handling of renewals/resubscriptions when the active subscription scheme no longer exists
See https://github.com/Prospress/woocommerce-subscribe-all-the-things/pull/136#discussion_r124692666

To be fixed:

- [x] General handling of cart items with an incorrectly applied scheme.
- [x] Handling of products with deleted scheme when re-subscribing.
- [x] Handling of products with deleted scheme when renewing.

franticpsyx:
Adds cart validation code to prevent items with incorrectly set/deleted schemes from being purchased.

In the simplest case, this could happen if the scheme data attached to a product changes _while_ the product is in a customer's cart.

The same logic is re-used to deal with re-subscriptions/initial payments of items with out-of-date scheme data.

The PR also includes a [patch](https://github.com/Prospress/woocommerce-subscribe-all-the-things/compare/2.0...issue-153#diff-ee624ed55e38fa547ae44e01c8c97a35R294) to allow renewals even when the originally subscribed scheme has been deleted.

Closes #153


includes/class-wcs-att-cart.php
```diff
@@ -129,6 +132,11 @@ public static function get_cart_subscription_schemes() {
 			if ( WCS_ATT_Product::is_subscription_product_type( $cart_item[ 'data' ] ) ) {
 				return false;
 			}
+
+			// Renewing/Resubscribing?
+			if ( isset( $cart_item[ 'subscription_renewal' ] ) || isset( $cart_item[ 'subscription_initial_payment' ] ) || isset( $cart_item[ 'subscription_resubscribe' ] ) ) {
+				return false;
```
franticpsyx: Not 100% sure about this.

I think I will need to test how cart subscriptions behave when re-subscribing -- with this in place I have the impression that cart-level schemes will not be appended to product-level schemes on cart item objects.

The reason I added this was that I wanted to prevent cart subscription options from showing up, but in hindsight this is probably a bad place to deal with this.

includes/class-wcs-att-cart.php
```diff
@@ -456,7 +504,63 @@ public static function is_supported_product_type( $arg ) {
...
+						// If not, show a blocking notice. Later on, we can add a button here to provide an opportunity to the user to resolve the problem.
+						} else {
+							$prompt = '';
+						}
+
+						wc_add_notice( sprintf( __( 'The &quot;%1$s&quot; subscription that you signed up for is no longer available. %2$s', 'woocommerce-subscribe-all-the-things' ), $cart_item[ 'data' ]->get_name(), $prompt ), 'error' );
 
+					// ...or a dev misbehaved and deserves some bad karma.
+					} else {
+						error_log( sprintf( 'Incorrect subscription scheme applied to cart item %s (%s). Scheme to apply: "%s". Applied scheme: "%s".' ), $cart_item_key, $cart_item[ 'data' ]->get_name(), var_export( $scheme_to_apply, true ), var_export( $applied_scheme, true ) );
```
thenbrent: PHP error logs can be hard to get a hold of. We like to use instances of `WC_Logger` to log things so that all we need to review the logs are temp admin creds to the site, instead of a server savvy admin with knowledge of where their PHP/WP install stores PHP error logs (if it even does). See [`WCS_Upgrade_Logger `](https://github.com/Prospress/woocommerce-subscriptions/blob/master/includes/upgrades/class-wcs-upgrade-logger.php) for an example.

The other option is to throw and exception to make sure that misbehaving/fishy dev is notified of their behaviour.

includes/class-wcs-att-cart.php
franticpsyx: @thenbrent this will prevent cart-level subscription schemes from being added and must be removed from here in order to allow these types of renewals/resubs. Not sure if this makes sense?

I need to fix this on the "displaying" side of things -- e.g. "when resubscribing/renewing, append the cart-level (global) schemes to objects but don't display any options in the cart page". Will see if I can patch it today 👍

includes/class-wcs-att-cart.php
```diff
@@ -242,9 +250,22 @@ public static function load_cart_item_data_from_session( $cart_item, $item_sessi
...
+				WCS_ATT_Product_Schemes::set_subscription_scheme( $cart_item[ 'data' ], $scheme_to_apply );
+
+				// Grab the applied scheme.
+				$applied_scheme = WCS_ATT_Product_Schemes::get_subscription_scheme( $cart_item[ 'data' ] );
+
+				// If the scheme was not applied sucessfully, then it was probably deleted, or something fishy happened.
+				if ( $scheme_to_apply !== $applied_scheme ) {
+					// In this case, simply ensure that no scheme is set on the object and handle the mismatch later.
+					WCS_ATT_Product_Schemes::set_subscription_scheme( $cart_item[ 'data' ], null );
+				}
```
franticpsyx: @thenbrent since we are validating on `WCS_ATT_Product_Schemes ::set_subscription_scheme` as well as on `WCS_ATT_Product_Schemes ::get_subscription_scheme`, I agree this is probably not needed! 👍

includes/class-wcs-att-cart.php
```diff
@@ -269,36 +290,63 @@ public static function apply_subscription_schemes( $cart ) {
...
-				}
+				/*
+				 * If renewing a subscription, delete all scheme-releted data from the object:
+				 * WCS has already linked it with a subscription order and will even copy the right total from there.
+				 */
+				if ( isset( $cart_item[ 'subscription_renewal' ] ) ) {
 
-				// Initialize subscription scheme data.
-				$cart_item[ 'wcsatt_data' ][ 'active_subscription_scheme' ] = self::get_subscription_scheme_to_apply( $cart_item );
+					WCS_ATT_Product_Schemes::set_subscription_schemes( $cart->cart_contents[ $cart_item_key ][ 'data' ], array() );
```
franticpsyx: Note there's an *s*, here we are removing all schemes on the object. This is because renewal cart items don't require any scheme data, as everything is copied over from the subscription object and handled by WCS.

danielbair:
Forgive me for hijacking this thread. I just want to vote for not breaking
cart level subscriptions. This is the only feature we use. We purchased WCS
thinking this was a standard function only to not find it there. Then, just
before we gave up, we discovered WCS_ATT and the cart subscription feature,
and now our donation website is working as desired.

Thanks,


-Daniel

includes/class-wcs-att-cart.php
franticpsyx: > PHP error logs can be hard to get a hold of.

So true. @thenbrent should we create a new log file for use by SATT? Or dump this into an existing WCS log file?

I am not sure if this is a "personal" thing -- I know many devs are developing with PHP/webserver error logs constantly visible. I would prefer to see an entry in this log when the application does something unexpected/forbidden that doesn't justify fatal-errorring. A filter callback function doesn't (?) feel like a good place to throw an Exception as it can't be caught.

franticpsyx:
@danielbair Thanks for chiming in! The last commit should fix this.

@thenbrent Thinking we should log incorrectly applied sub schemes only when an order is created. We should keep our log files as clean as possible.

franticpsyx:
@danielbair Actually we need more people to "hijack" these threads like this to improve SATT 😄 

Again, thanks for participating!

includes/class-wcs-att-cart.php
franticpsyx: Took another look, I think it's needed: 

Example -- the "scheme to apply" might not match the one that's actually applied on the object, but both might be valid. Here we are simply detecting the mismatch and resetting the "applied one" to `null`.

includes/class-wcs-att-cart.php
franticpsyx: @thenbrent Added some basic logging at https://github.com/Prospress/woocommerce-subscribe-all-the-things/pull/173/files#diff-e5bc18f1662c7e2ec4e4533bf5f1dac3R149

Note we don't log the Order / Item ID there because the CRUD object hasn't been saved yet. I thought the timestamp would be enough for any future investigations, let me know if you think otherwise.

franticpsyx:
Should be good to go.

PR #199, opened 2017-08-24
Fix prompt display for variable products

### LINKED ISSUE #198: Custom subscription prompt doesn't show on variable product
This one's pretty self-explanatory. 
- Add and save a custom prompt for a variable product
- On the single product page, select variation options
- See that the default 'Choose a subscription' prompt is displayed instead of your custom text

Looks like the _wcsatt_subscription_prompt meta call is made on the variation, so it's not returning the meta that was saved on the parent product.

willgorham:
Fixes #198 

Open to feedback here. Still learning how WC handles things... :)

### DIFF
--- a/includes/display/class-wcs-att-display-product.php
+++ b/includes/display/class-wcs-att-display-product.php
@@ -55,7 +55,7 @@ class WCS_ATT_Display_Product {
 	 * @param  WC_Product  $product
 	 * @return void
 	 */
-	public static function get_subscription_options_content( $product ) {
+	public static function get_subscription_options_content( $product, $variable_product = null ) {
 
 		$content = '';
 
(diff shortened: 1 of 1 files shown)


willgorham:
Fixes #198

franticpsyx:
@willgorham thanks for this 👍 

Looks good. I think the best way to handle this in the long run would be to include a new argument, as you did. I think WooCommerce 2 made it possible to retrieve the parent from the variation, but this is no longer the case in 3.0.

One change I would suggest would be to rename `$variable_product` to `$product_parent`. A type-agnostic name choice there could come handy later. Happy to merge this otherwise!

willgorham:
@franticpsyx Makes sense to me - updated to `$parent_product` 👍

Issue #26, opened 2016-03-28
Handling of scheme ids


franticpsyx:
Ref - https://github.com/Prospress/woocommerce-subscribe-to-all-the-things/pull/25

@thenbrent Opening this to discuss about the way scheme IDs are handled. I feel this might be a bit of an annoyance for the active user base:

>this requires product level subscription schemes to be resaved to continue to work

Also, it might be wise to keep the door open for dynamically modifying the id later on, in case we add extra settings that need to find their way into the id (sync?).

Adding to the 1.1 milestone.

franticpsyx:
Changed options arrays to use id-based keys in template files.

Left all other code intact as explained above, on standby for possible bigger changes later on. Closing for now.

PR #218, opened 2017-10-20
Added missing classes for item options in cart.

seb86:


franticpsyx:
@seb86 are these "missing" in the sense they existed in v1?

franticpsyx:
NM, looks like you are referring to the product template 👍

franticpsyx:
I don't think there's a `value` key there so adding this in would likely trigger a warning -- looks like a mild inconsistency between product/cart templates that we can/should fix in a BW compatible manner. Closing.

seb86:
I did not see an error when I tested this change. Appears to be fine unless you can confirm otherwise. You also have `$option[ 'value' ]` used on [line #26](https://github.com/seb86/woocommerce-subscribe-to-all-the-things/blob/5579d71ee7f5259eb8b308a2ee7bfa417c019f1b/templates/cart/cart-item-subscription-options.php#L26) for the input. `value="<?php echo esc_attr( $option[ 'value' ] ); ?>"`

franticpsyx:
Pulled and checked again, this looks in order.

Issue #227, opened 2017-11-05
Enhancement: Allow filtering of scheme discount


willgorham:
### Use Case
I have a shop where SATT products are configured with inherited prices and a percent discount. The shop has a 'wholesaler' user role. These users are already set up to get bulk product discounts for standard one-time purchases (via the Dynamic Pricing extension), and I would like to not give them the additional subscription discount on top of that if they make a subscription purchase. 

### Potential Solution
There might be a better way to go about this, so this is a bit of a question, but I'll send over a PR and see what the feedback is or if there's a better option to make this happen. A filter on the scheme discount (specifically in `WCS_ATT_Scheme->get_discount()`) would allow me to easily change the discount based on the current user role.

--
PR: #228

seb86:
@willgorham With this filter, are you able to detect which scheme the discount is applied too?

Can you give an example of the filter in use?

willgorham:
@seb86 

You can't detect the scheme at the moment, no. Which I suppose is pretty useless if you need any context. My current use is simply:
```
add_filter( 'wcsatt_subscription_scheme_discount', 'bbe_wcsatt_subscription_scheme_discount', 10, 1 );
function bbe_wcsatt_subscription_scheme_discount( $discount ) {

  // Exclude user roles
  if ( current_user_can( 'reseller' ) ||
       current_user_can( 'wholesaler' ) ||
       current_user_can( 'retailer' ) ) {
    $discount = 0;
  }

  return $discount;
}
```

It would be more useful I suppose if I passed `$this->data` or even just `$this` into the filter as an argument. What do you think?

seb86:
I think an argument may be required otherwise it will apply to all products that have subscription schemes.

What do you think @franticpsyx?

willgorham:
Right. In my use case I'm applying this to all products. I was trying to keep things as simple as possible - not adding too much functionality if it's not needed. But if you want to make this useful in other cases and contexts then an additional argument makes sense.

franticpsyx:
@willgorham -- how about `wcsatt_price_filters_allowed`?

In general I'd say that if your intent is to work with prices / discount, the best place to look / modify would be the `WCS_ATT_Product_Prices` class.

The `WCS_ATT_Scheme` class is merely a data abstraction -- it does not contain anything to help you identify the product instance that the scheme is attached to, which is common requirement when working with discounts.

[...]

PR #228, opened 2017-11-05
Add filter to scheme discount

willgorham:
Potential solution for #227.

This filter provides the ability to adjust the scheme discount on the fly, for example, based on the current user's role.


willgorham:
No longer needed - resolved #227 using existing codebase.

PR #237, opened 2017-12-07
Scheme Switching for Core Types

### LINKED ISSUE #233: WCS Features Parity Milestone 3: Scheme Switching for Core Types
From https://github.com/Prospress/woocommerce-subscribe-all-the-things/wiki/Roadmap

franticpsyx:
Closes #233 

Ran some simple test cases with Simple + Variable products, seems to work. 

I'd love to have an extra confirmation, though, preferably from someone who knows how things can break in WCS. I saw some open issues related to switching in https://github.com/Prospress/woocommerce-subscriptions/issues/2452 .

Since we haven't added syncing to SATT yet, I didn't consider testing anything beyond confirming that schemes + attributes can be switched successfully. So I haven't done any testing to confirm that proration works as intended here. Not sure if I should.

Note that this enables switching for all products with subscription schemes via `wcs_is_product_switchable`. Should we have a global option to disable this somewhere -- perhaps around here: https://cl.ly/450l2E392o01 ?

Ideally I think switching should be enabled/disabled at product level, though. SATT might be a good opportunity to deviate from WCS 1/2 if it's something we _want_ for WCS 3+.

### DIFF
--- a/includes/class-wcs-att-cart.php
+++ b/includes/class-wcs-att-cart.php
@@ -224,7 +224,6 @@ class WCS_ATT_Cart {
 		return $cart_item;
 	}
 
-
 	/**
 	 * Load saved session data of cart items that can be pruchased on a recurring basis.
 	 *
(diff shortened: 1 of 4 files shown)


franticpsyx:
> Note that this enables switching for all products with subscription schemes via wcs_is_product_switchable. Should we have a global option to disable this somewhere -- perhaps around here: https://cl.ly/450l2E392o01 ?

> Ideally I think switching should be enabled/disabled at product level, though. SATT might be a good opportunity to deviate from WCS 1/2 if it's something we want for WCS 3+.

@thenbrent would love your opinion about this -- always thought that switching should be controlled at product level.

Issue #262, opened 2018-01-30
Subscription Management: Add entire carts to existing Subscriptions


franticpsyx:
Branched off from #239 .

franticpsyx:
This was an interesting to look at, partly because of the need to design a flow (see https://cl.ly/1W1O3l133w1Q) and partly because it made me realize that the solution to #239 and #240 could be improved.

If you look at https://github.com/Prospress/woocommerce-subscribe-all-the-things/pull/252 you'll notice that when a product is added to an existing subscription, it is _not_ added to the cart first, which means that any custom logic that runs in cart context will not be given an opportunity to run.

The need to work at cart level to solve this one gave me the idea to re-use the same add-cart-to-subscription logic to re-work #239 and #240 .

What happens now is that when adding a product to a subscription:

- Existing cart contents are backed up.
- The item(s) associated with the product we want to add to a subscription are added to the cart and totals calculated.
- We iterate over (recurring) cart contents and add them to the selected subscription, making sure that cart totals are maintained correctly.
- We reset the cart to its initial state.

This allows us to be more precise with our implementation, and keeps the codebase as dry as possible. The added complexity could be a cause for concern (plugin conflicts) compared to the "old" (a couple weeks old?) way, but the new approach is one that we'd likely need to go for down the road, as I suspect we'd be faced with... plugin compatibility issues otherwise (plugins conditionally modifying cart prices, mostly).

Another thing I noticed is that the `WCS_ATT_Add` management class grew too big for my liking, so I'd like to break things up there a little before we declare `2.1` done: https://github.com/Prospress/woocommerce-subscribe-all-the-things/issues/274

franticpsyx:
Closed via #275

Issue #274, opened 2018-02-12
Subscription Management: Refactor `WCS_ATT_Add` class


franticpsyx:
Needs to be broken up in 3 parts:

- Application layer (the 'add-to-subscription' API)
- Hooks specific to the add-product case
- Hooks specific to the add-cart case

franticpsyx:
Closed via 92136a33665ce2aeb2a0caaea7cc90a4a0826bfe

PR #275, opened 2018-02-12
Add carts to subscriptions

franticpsyx:
See #262

This shamelessly large commit includes 2 things:

* Introduces add-cart-to-subscription functionality, at the core of which is `WCS_ATT_Add::add_cart_to_subscription`.
* Refactors existing add-product-subscription functionality to re-use `WCS_ATT_Add::add_cart_to_subscription`.

From #262 --

```
What happens now is that when adding a product to a subscription:

- Existing cart contents are backed up.
- The item(s) associated with the product we want to add to a subscription are added to the cart and totals calculated.
- We iterate over (recurring) cart contents and add them to the selected subscription, making sure that cart totals are maintained correctly.
- We reset the cart to its initial state.
```


franticpsyx:
Merging.

I'd suggest we review the entire `2.1` change set. Although it may seem counter-productive now, the 2.1 codebase should be merely "sitting on top of 2.0", which _should_ make it fairly easy to tell what's been added and why.

Doing this will allow us to look at 2.1 in a more holistic manner, which is always a good idea when working on UI/UX-intensive features.

Finally, we need to deal with #274 before doing a code review.

PR #302, opened 2018-05-10
Wrong url/text-domain in .pot file

jrick1229:
When using the .pot file, the domain at the top for issues is listing the wrong url.
This isn't a huge deal, but I was unfortunately using it for the text-domain field for translations. It wasn't working, then I realized why.


franticpsyx:
Ops! Thanks so much for taking the time to report this @jrick1229 --

This is generated by: https://github.com/Prospress/woocommerce-subscribe-all-the-things/blob/master/Gruntfile.js#L86 -- so it needs to be edited there to ensure that the .pot file headers will be correct after the next build process runs.

I also found the same error [here](https://github.com/Prospress/woocommerce-subscribe-all-the-things/blob/master/package.json#L6) and [here](https://github.com/Prospress/woocommerce-subscribe-all-the-things/blob/master/package.json#L9)

Would be grateful if you could add these 3 lines to this PR. Thanks so much!

PS: "Subscribe To All The Things" is the correct name (grammatically) but I've always found the (wrong) "Subscribe All The Things" more direct and fun.

jrick1229:
Surely, @franticpsyx!
I fixed that, but can you let me know what I need to do about the Gruntfile? (not so used to this on GitHub still, learned on Bitbucket)
Thanks!

franticpsyx:
Hey @jrick1229 :) If you look at https://github.com/Prospress/woocommerce-subscribe-all-the-things/blob/master/Gruntfile.js#L86 you'll see there's another incorrect reference to this repo. You don't need to do anything beyond editing/fixing that reference.

The `.pot` file is generated by the build process so if you patch that JS file the `.pot` file will be automagically patched (no need to be concerned about this at all).

franticpsyx:
All done here, thanks for helping with this @jrick1229 :)

PR #360, opened 2018-10-16
add info on where to get zip file

tansaku:
the installation instructions are a little confusing since step 4 refers to a zip in step 1 that's not actually mentioned ...

### DIFF
--- a/README.md
+++ b/README.md
@@ -51,10 +51,11 @@ To use the plugin, you will need:
 
 ### Installation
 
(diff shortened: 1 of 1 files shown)


thenbrent:
Thanks @tansaku. I've addressed this with SHA: b4aad07 - I used to a separate commit to use [link tags for the zip](https://github.com/Prospress/woocommerce-subscribe-all-the-things/archive/master.zip) instead of a plain text full URL.

PR #366, opened 2018-10-21
Resolve conflict between NYP variations and SATT. 

helgatheviking:
Displays both NYP price input and SATT options (first pass)

Problem:

If a variable product has an NYP variation, then the SATT options do not display:

image

This is because NYP also filters `woocommerce_available_variation` to change the `price_html` [here](https://github.com/woocommerce/woocommerce-name-your-price/blob/master/includes/class-wc-name-your-price-display.php#L408).

By avoiding this filter and displaying the SATT scheme options in the same before add to cart button location, the two are able to display simultaneously.

image

Given that current the SATT options are not variation-specific, it seems unnecessary to attach them to each variation, but I may not understand that reasoning yet.

### DIFF
--- a/assets/css/wcs-att-frontend.css
+++ b/assets/css/wcs-att-frontend.css
@@ -42,6 +42,10 @@ ul.wcsatt-options-product {
 	display: inline;
 }
 
+.woocommerce-variation-add-to-cart .wcsatt-options-wrapper {
+	display: none;
+}
+
 /* Cart */
 
 ul.wcsatt-options-cart {
--- a/assets/js/wcs-att-single-add-to-cart.js
+++ b/assets/js/wcs-att-single-add-to-cart.js
@@ -91,6 +91,9 @@
 			variation_found: function( event, variation ) {
 				this.variation = variation;
 				this.initialize( { $el_options: this.$el.find( '.wcsatt-options-wrapper' ) } );
+
+
+				this.$el.find( '.wcsatt-options-wrapper' ).slideDown();
 			},
 
 			variation_selected: function() {
@@ -98,6 +101,9 @@
 			},
 
 			reset_schemes: function() {
+
+				this.$el.find( '.wcsatt-options-wrapper' ).slideUp();
+
 				this.variation = false;
 				this.model.set_schemes( {} );
 				this.model.set_active_scheme( false );
@@ -122,7 +128,6 @@
 			},
 
 			initialize: function( options ) {
-
 				this.$el_options = options.$el_options;
 
 				this.model.set_schemes( this.find_schemes() );
@@ -736,6 +741,13 @@
 	$( '.composite_form .composite_data' ).each( function() {
 		$( this ).on( 'wc-composite-initializing', function( event, composite ) {
 			new CP_Integration( composite );
+		} );
+	} );
+
+	// Hook into Mix and Match.
+	$( '.mnm_form .mnm_cart' ).each( function() {
+		$( this ).on( 'wc-mnm-initialized', function( event, mnm ) {
+			new MNM_Integration( bundle );
 		} );
 	} );
 
--- a/includes/display/class-wcs-att-display-product.php
+++ b/includes/display/class-wcs-att-display-product.php
@@ -45,8 +45,6 @@ class WCS_ATT_Display_Product {
 		add_filter( 'woocommerce_product_add_to_cart_url', array( __CLASS__, 'add_to_cart_url' ), 10, 2 );
 		add_filter( 'woocommerce_product_supports', array( __CLASS__, 'supports_ajax_add_to_cart' ), 10, 3 );
 
-		// Replace plain variation price html with subscription options template.
-		add_filter( 'woocommerce_available_variation', array( __CLASS__, 'add_subscription_options_to_variation_data' ), 0, 3 );
 	}
 
 	/**
@@ -59,13 +57,6 @@ class WCS_ATT_Display_Product {
 	public static function get_subscription_options_content( $product, $parent_product = null ) {
 
 		if ( ! WCS_ATT_Product::supports_feature( $product, 'subscription_scheme_options_product_single' ) ) {
-			return '';
-		}
-
-		/*
-		 * Subscription options for variable products are embedded inside the variation data 'price_html' field and updated by the core variations script.
-		 */
-		if ( $product->is_type( 'variable' ) ) {
 			return '';
 		}
 


franticpsyx:
Hey @helgatheviking thanks for looking into this!

As mentioned on Slack, each variation will eventually be able to support its own schemes, too -- so letting WC handle the "view" updates via variation changes is the most obvious way to go.

Otherwise, SATT will need to update its own view with each variation change -- that's introducing an overhead I'm not happy with.

However, this is already a problem _now_:

Without testing this PR in detail, I don't think it's going to work with scheme-level discounts: Each variation may have its own price. When there's a need to show a different before/after scheme price for each variation (scheme discount present), a dedicated JS view will be needed to render scheme options when variations change.

Arguably you could do the same in NYP (take the view out of the variation html) with less code.

franticpsyx:
Hey @helgatheviking thanks for looking into this!

As mentioned on Slack, each variation will eventually be able to support its own schemes, too -- so letting WC handle the "view" updates via variation changes is the most obvious way to go.

Otherwise, SATT will need to update its own view with each variation change -- that's introducing an overhead I'm not happy with.

However, this is already a problem _now_:

Without testing this PR in detail, I don't think it's going to work with scheme-level discounts: Each variation may have its own price. When there's a need to show a different before/after scheme price for each variation (scheme discount present), a dedicated JS view will be needed to render scheme options when variations change.

Arguably you could do the same in NYP (take the view out of the variation html) with less code.

franticpsyx:
Hey @helgatheviking thanks for looking into this!

As mentioned on Slack, each variation will eventually be able to support its own schemes, too -- so letting WC handle the "view" updates via variation changes is the most obvious way to go.

Otherwise, SATT will need to update its own view with each variation change -- that's introducing an overhead I'm not happy with.

However, this is already a problem _now_:

Without testing this PR in detail, I don't think it's going to work with scheme-level discounts: Each variation may have its own price. When there's a need to show a different before/after scheme price for each variation (scheme discount present), a dedicated JS view will be needed to render scheme options when variations change.

Arguably you could do the same in NYP (take the view out of the variation html) with less code.

franticpsyx:
Hey @helgatheviking thanks for looking into this!

As mentioned on Slack, each variation will eventually be able to support its own schemes, too -- so letting WC handle the "view" updates via variation changes is the most obvious way to go.

Otherwise, SATT will need to update its own view with each variation change -- that's introducing an overhead I'm not happy with.

However, this is already a problem _now_:

Without testing this PR in detail, I don't think it's going to work with scheme-level discounts: Each variation may have its own price. When there's a need to show a different before/after scheme price for each variation (scheme discount present), a dedicated JS view will be needed to render scheme options when variations change.

Arguably you could do the same in NYP (take the view out of the variation html) with less code.

franticpsyx:
Hey @helgatheviking thanks for looking into this!

As mentioned on Slack, each variation will eventually be able to support its own schemes, too -- so letting WC handle the "view" updates via variation changes is the most obvious way to go.

Otherwise, SATT will need to update its own view with each variation change -- that's introducing an overhead I'm not happy with.

However, this is already a problem _now_:

Without testing this PR in detail, I don't think it's going to work with scheme-level discounts: Each variation may have its own price. When there's a need to show a different before/after scheme price for each variation (scheme discount present), a dedicated JS view will be needed to render scheme options when variations change.

Arguably you could do the same in NYP (take the view out of the variation html) with less code.

franticpsyx:
Hey @helgatheviking thanks for looking into this!

As mentioned on Slack, each variation will eventually be able to support its own schemes, too -- so letting WC handle the "view" updates via variation changes is the most obvious way to go.

Otherwise, SATT will need to update its own view with each variation change -- that's introducing an overhead I'm not happy with.

However, this is already a problem _now_:

Without testing this PR in detail, I don't think it's going to work with scheme-level discounts: Each variation may have its own price. When there's a need to show a different before/after scheme price for each variation (scheme discount present), a dedicated JS view will be needed to render scheme options when variations change.

Arguably you could do the same in NYP (take the view out of the variation html) with less code.

franticpsyx:
Hey @helgatheviking thanks for looking into this!

As mentioned on Slack, each variation will eventually be able to support its own schemes, too -- so letting WC handle the "view" updates via variation changes is the most obvious way to go.

Otherwise, SATT will need to update its own view with each variation change -- that's introducing an overhead I'm not happy with.

However, this is already a problem _now_:

Without testing this PR in detail, I don't think it's going to work with scheme-level discounts: Each variation may have its own price. When there's a need to show a different before/after scheme price for each variation (scheme discount present), a dedicated JS view will be needed to render scheme options when variations change.

Arguably you could do the same in NYP (take the view out of the variation html) with less code.

franticpsyx:
Hey @helgatheviking thanks for looking into this!

As mentioned on Slack, each variation will eventually be able to support its own schemes, too -- so letting WC handle the "view" updates via variation changes is the most obvious way to go.

Otherwise, SATT will need to update its own view with each variation change -- that's introducing an overhead I'm not happy with.

However, this is already a problem _now_:

Without testing this PR in detail, I don't think it's going to work with scheme-level discounts: Each variation may have its own price. When there's a need to show a different before/after scheme price for each variation (scheme discount present), a dedicated JS view will be needed to render scheme options when variations change.

Arguably you could do the same in NYP (take the view out of the variation html) with less code.

franticpsyx:
Hey Kathy, thanks for looking into this!

As mentioned on Slack, each variation will eventually be able to support its own schemes, too -- so letting WC handle the "view" updates via variation changes is the most obvious way to go.

Otherwise, SATT will need to update its own view with each variation change -- that's introducing an overhead I'm not happy with.

However, this is already a problem _now_:

Without testing this PR in detail, I don't think it's going to work with scheme-level discounts: Each variation may have its own price. When there's a need to show a different before/after scheme price for each variation (scheme discount present), a dedicated JS view will be needed to render scheme options when variations change.

Arguably you could do the same in NYP (take the view out of the variation html) with less code.

franticpsyx:
Hey @helgatheviking thanks for looking into this!

As mentioned on Slack, each variation will eventually be able to support its own schemes, too -- so letting WC handle the "view" updates via variation changes is the most obvious way to go.

Otherwise, SATT will need to update its own view with each variation change -- that's introducing an overhead I'm not happy with.

However, this is already a problem _now_:

Without testing this PR in detail, I don't think it's going to work with scheme-level discounts: Each variation may have its own price. When there's a need to show a different before/after scheme price for each variation (scheme discount present), a dedicated JS view will be needed to render scheme options when variations change.

Arguably you could do the same in NYP (take the view out of the variation html) with less code.

helgatheviking:
Hi Manos! This is some good feedback and provides some info that I didn't know about SATT, which hopefully we guide me towards a better solution. 

>so letting WC handle the "view" updates via variation changes is the most obvious way to go.

That makes sense, but currently there aren't any hooks in https://github.com/woocommerce/woocommerce/blob/master/templates/single-product/add-to-cart/variation.php for adding additional blobs of info automatically. Even if I sent a PR for that and they liked it, it wouldn't be released for a bit. 

However, I think it should be possible to add a different key to the `woocommerce_available_variation` array and insert that into the `<div class=woocommerce-variation single_variation"></div>` on the `found_variation` trigger (or maybe there's even another earlier one for when the template is rendered.... I'll need to investigate that further. But doing everything on the `price_html` key is what is conflicting with NYP, which is also modifying the `price_html`.

>Without testing this PR in detail, I don't think it's going to work with scheme-level discounts: Each variation may have its own price. When there's a need to show a different before/after scheme price for each variation (scheme discount present), a dedicated JS view will be needed to render scheme options when variations change.

I definitely didn't test in detail either! Kinda glad I didn't since it's obvious to you it won't work as is. and NYP variation probably doesn't seem to support a discount, but it can't break the discounts in the other variations. 

>Arguably you could do the same in NYP (take the view out of the variation html) with less code.

Idk yet if it's less code. I only need 1 input. Though I suppose it would reduce the amount of JS code I need to hide/show said input and also reducing the amount of JS needed to modify the min/max/error data attributes, etc. Crap, my NYP 3.0 list is already really long... did you just add to it? ;)

helgatheviking:
Moving these comments here for reference:
>NYP does support different suggested/min prices per variation, _but_ the input is printed on `woocommerce_single_variation` once and it's visibility is toggled on the `found_variation` event.

>NYP is filtering the variation `price_html` to switch it to be the suggested price, ex "Suggested Price: $50"

helgatheviking:
The price input is still in the wrong place, but check out this commit for switching the satt options out of the `price_html` key in the available variation array. https://github.com/helgatheviking/woocommerce-subscribe-all-the-things/commit/ab3709090b7e6774b59a30eca0ea92766675a956 This keeps NYP from stomping on the satt options, while still letting Woo handle the heavy lifting. 

I will need to dabble more with showing the NYP input this way too.

franticpsyx:
@helgatheviking this could be a possible solution, but still wouldn't be enough. What you want to do is make sure that SATT doesn't display its own price html strings. I think you touched upon it in #369 ?

I think we should open a "Support NYP" issue and summarize everything there. If this requires changes in SATT I'd suggest adding this to the 2.2 milestone.

helgatheviking:
Closing in favor of #371

PR #373, opened 2018-10-24
Variable nyp compat 2

### LINKED ISSUE #371: NYP Compatibility
- [x] If a variable product has an NYP variation, then the SATT options do not display: This is because NYP also filters `woocommerce_available_variation` to change the `price_html` [here]. [Propose adding SATT as different key](ab3709090b7e6774b59a30eca0ea92766675a956)
- [x] If no suggested price (the price string ends up null) the subscription terms for a scheme are blank. [Propose filtering args](0d9a275f4e04b5e89fd566068ac427c69d3772b7)

Once you get both to display you SATT options and NYP price input to display you still have two issues:
- [x] Scheme Discount is applied to NYP price. If you enter $1234 as the price is $1110 in the cart with a 10% discount. For NYP, 
- [x] SATT options appear before NYP input. 
- [x] If suggested price exists, then it shows up in the SATT options

**SATT Options Do Not Display**
image

**If price string is null (ie: no suggested price for NYP), terms do not display**
image

**Suggested price string in the SATT options**
image

helgatheviking:
Move the SATT options output to `woocommerce_single_variation hook` hook, but still dynamically load the content on the `variation_found` JS trigger... fixes display order with NYP.

Filter subscription price string args to allow NYP to manipulate strings.

Closes #371

### DIFF
--- a/assets/js/wcs-att-single-add-to-cart.js
+++ b/assets/js/wcs-att-single-add-to-cart.js
@@ -90,6 +90,7 @@
 
 			variation_found: function( event, variation ) {
 				this.variation = variation;
+				this.$el.find( '.single_variation' ).append( variation.satt_html );
 				this.initialize( { $el_options: this.$el.find( '.wcsatt-options-wrapper' ) } );
 			},
 
--- a/assets/js/wcs-att-single-add-to-cart.js
+++ b/assets/js/wcs-att-single-add-to-cart.js
@@ -90,7 +90,7 @@
 
 			variation_found: function( event, variation ) {
 				this.variation = variation;
-				this.$el.find( '.single_variation' ).append( variation.satt_html );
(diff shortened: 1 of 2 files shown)


franticpsyx:
@helgatheviking 

I'm a bit skeptical about including this in a fix release as there are potentially breaking changes here. Another concern is that we'd need to put some time into testing all affected cases thoroughly. The resources we have available right now are very limited (SATT is still a side-project), which is why we can only include only small patches/fixes in fix releases.

This is a small architectural change that may potentially require a second look at short notice -- and because we can't be very thorough with testing at this point, I'd rather delay this one for 2.2.

franticpsyx:
@helgatheviking closing this one, let's chat about MnM support in APFS on Slack.

Issue #396, opened 2018-12-26
How do I use this plugin with Woocommerce Memberships


boodaah:
This plugin works great for a subscribe and save feature, but I have the membership plugin where members get a 15% discount off. In combination with this plugin I see issues.
1) The "- or subscribe and get 10% off" is duplicated
2) The subscription plan prices get a compounded discount, both the subscribe discount and the membership discount.

Eg.
![Example](https://user-images.githubusercontent.com/3849292/50453829-5401aa00-0911-11e9-9ecc-77e8815ff9ba.png)

What can I do to get the prices to 10% off full price when subscribed, or 15% off full price when they're a member?

rrennick:
Check to see if your theme has template overrides.

boodaah:
The pull request addresses my first issue about the duplicated suffix.

My second issue is an edge case issue. My ugly fix was to set the subscription_discount = 0 in the constructor of includes/class-wcs-att-scheme.php if the user was going to get a membership discount. Basically, active members don't get a subscription discount.

```
		if ( function_exists( 'wc_memberships' ) ) {  
			if ( wc_memberships_is_user_active_member() ) {  
				$args[ 'data' ][ 'subscription_discount' ] = 0;  
			}  
		}  
```

I doubt there is any appeal to modify the base code for a case like this, but I thought I'd share.

franticpsyx:
Hey @boodaah 

Due to the amount of dependencies (SATT is a complex extension of Subscriptions, which is a complex extension of WooCommerce), we have to be very picky about the extensions/plugins we support in SATT. WooCommerce Memberships is not officially supported on our side at this point.

There are likely more edge-case issues/bugs that will need to be sorted once we officially support Memberships, so I'm currently skeptical about including patches that deal with some isolated symptoms. I'd love follow a more systematic approach here and eventually declare Memberships officially compatible with SATT, but this isn't anywhere close to being planned yet.

I understand solving this is probably vital for you, but we _must_ be very systematic when it comes to making changes to a project with such a heavy dependencies overhead.

PR #398, opened 2018-12-28
Prevent duplicate suffix on price html

boodaah:
#396 I found that the price suffix would be duplicated when the plugin is used with other plugins that give price discounts.

### DIFF
--- a/includes/product/class-wcs-att-product-prices.php
+++ b/includes/product/class-wcs-att-product-prices.php
@@ -241,7 +241,9 @@ class WCS_ATT_Product_Prices {
 						}
 					}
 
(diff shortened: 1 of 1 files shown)


franticpsyx:
Hey thanks for contributing @boodaah 

This addresses the symptom, not the cause. Having worked with the Memberships discount class in the past I think I can guess why this this might be happening, but I don't immediately see a way to fix this that addresses the root cause. This should be part of a more in-depth look at adding Memberships support to SATT.

Your PR also comes at a turning point for SATT, as we are actively looking at changing its distribution/support model _precisely_ in order to be able to continue maintaining and improving it in a more systematic and sustainable manner. Details to follow soon.

PR #428, opened 2019-02-18
alters response json on subscription scheme change

DragosMocrii:
This adds the possibility to alter the json object returned when the subscription scheme is changed in the cart. Also, triggers an event on the body element when the response is successful.

I needed this to display custom notices when the subscription scheme changes and some criteria where met.


franticpsyx:
Hey @DragosMocrii thanks for the PR!

Got a few objections to this.

1. You can already listen to `wcsatt_updated_cart` on `body` as JS events propagate upwards in the DOM.

2. I'm skeptical about adding `old_scheme` and `new_scheme` details in the response, as it's not needed by SATT in any way. If you application needs this information, it might be best to track changes in your own script.

